PR #8, opened 2019-05-03
Fs apr19 vitalii duma

dvitalijb:


src/style.css
```diff
@@ -0,0 +1,112 @@
...
+
+@media(max-width: 600px) {
+  div {
+    width: 300px;
+  }
+}
+
+@media(max-width: 450px) {
+  div {
+    width: 450px;
```
gestap: try to use `flex-basis`

PR #12, opened 2019-05-03
not ready, need TA's help to solve this

valeriitovstyk:


src/style.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.block {
+  min-width: 300px;
+  height: 300px;
+}
+
+@media (min-width: 600px) {
+  .block {
+    width: 50%;
```
YuriiHoliuk: will be better to use flex-grow with flex

PR #16, opened 2019-05-04
added page with the flexible six snake-colored boxes

YojikYo:


src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  font-family: Arial, Helvetica, sans-serif;
+  line-height: 300px;
+  font-size: 100px;
+  text-align: center;
+  color: #fff;
+}
+
+.block_1 {
+  background: hsl(0, 100%, 50%);
+  order: 1;
```
YuriiHoliuk: do not need to specify default order

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  order: 5;
+  background: hsl(0, 100%, 10%);
+}
+
+.block_6 {
+  order: 6;
+  background: hsl(0, 100%, 0%);
+}
+
+@media all and (min-width: 0) {
```
YuriiHoliuk: not needed, add to common styles

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+@media all and (min-width: 600px) {
+  .flex_container {
+    flex-flow: wrap;
+  }
+
+  .snake {
+    width: 50%;
```
YuriiHoliuk: try to use flex-basis instead of width

YojikYo: well, done

src/style.css
YojikYo: ok!

PR #22, opened 2019-05-06
Vitalii Pshychenko

VitaliiPshychenko:


src/index.html
```diff
@@ -13,7 +13,14 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+<section class="blocks-box">
+  <div class="block first"><span>1</span></div>
```
YuriiHoliuk: do not need additional span

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  height: 300px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  font-size: 100px;
+}
+
+.first {
+  background: rgb(255, 0, 0);
+  order: 1;
```
YuriiHoliuk: do not need to specify default order

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+
+  .second {
+    order: 2;
+  }
+
+  .third {
+    order: 3;
+  }
+
+  .fourth {
```
gestap: in this @media enough to add only order for `.fourth`

PR #23, opened 2019-05-06
solution

ilovepinkpony1:


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.sixth {
+  background-color: hsl(0, 100%, 0%);
+}
+
+@media (min-width: 600px) {
+  .box > div {
+	
+    width: 50%;
```
YuriiHoliuk: use flex-basis instead of width

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  background-color: hsl(0, 100%, 10%);
+}
+
+.sixth {
+  background-color: hsl(0, 100%, 0%);
+}
+
+@media (min-width: 600px) {
+  .box > div {
+
```
YuriiHoliuk: remove

src/style.css
```diff
@@ -0,0 +1,84 @@
+* {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
```
gestap: no need to use `*` for `font-family`
Add it to `body` or another class

PR #26, opened 2019-05-07
Fs apr19 roman o

ronyshchenko:


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  background-color: hsl(0, 100%, 10%);
+}
+
+.six {
+  background-color: hsl(0, 100%, 0%);
+}
+
+@media (min-width: 600px) {
+  .container > div {
+    height: 300px;
```
YuriiHoliuk: do not need to repeat height

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+    width: 50%;
```
YuriiHoliuk: also, use flex-basis instead of width

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  .six {
+    order: 6;
+  }
+}
+  
+@media (min-width: 900px) {
+  .container > div {
+    height: 300px;
```
nicholas38466: дубль

PR #70, opened 2019-06-03
Solution - Vitalik Puzko

mpvitalik:
- [DEMO LINK](https://mpvitalik.github.io/layout_snake/)
- [TEST REPORT LINK](https://mpvitalik.github.io/layout_snake/report/html_report/)


mgrinko:
По заданию ширина не может быть меньше высоты
> so that the width of each block is not less than its height (300px).

PR #100, opened 2019-07-26
Fs jul19 konstantin b

burnazaki:
[https://burnazaki.github.io/layout_snake/](DEMO)
[https://burnazaki.github.io/layout_snake/report/html_report/](TEST REPORT)


src/style.css
```diff
@@ -0,0 +1,94 @@
+* {
```
irenhh: do not overuse `*`-selector, style the elements itself,
if you want to reset body margins, just style body tag

src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  }
+  .sixth {
+    order: 4;
+  }
+}
+
+@media all and (min-width: 1200px) {
+  .flex-block {
+    min-width: 400px;
+    flex-grow: unset;
```
irenhh: you don't need flex-grow here

src/style.css
```diff
@@ -0,0 +1,93 @@
+body {
+  margin: 0;
+  padding: 0;
```
irenhh: body doesn't have default paddings

PR #106, opened 2019-07-26
Vika Tsykun solution

ViktoriiaTsykun:
- [DEMO LINK](https://viktoriiatsykun.github.io/layout_snake/)
- [TEST REPORT LINK](https://viktoriiatsykun.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,74 @@
+* {
+  margin: 0;
+}
+
+.snake {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,74 @@
+* {
```
irenhh: reset body margins with body styling

ViktoriiaTsykun: fixed

PR #108, opened 2019-07-26
My solution for 'Snake'

VladyslavSak:
On my mind. we can add to width (33%) in the different media queries for the more adaptive site, but the CSS-test doesn`t allow do it.


irenhh:
you didn't publish last changes

VladyslavSak:
I did deploy and publish my changes.

PR #110, opened 2019-07-26
Fs jul19 oleksii demchenko

alexDemche:
* [Demo link](https://alexdemche.github.io/layout_snake/)
* [Test report](https://alexdemche.github.io/layout_snake//report/html_report/)

Please, can you help me after you check out my work, i am newbie in BEM naming.
Did i do it right ?

Or should I name the classes as in the example below ?
What will be more better solution for this particular task ?
```
<div class="boxes">
  <div class="boxes__box-1">1</div>
  <div class="boxes__box-2">2</div>
  <div class="boxes__box-3">3</div>
  <div class="boxes__box-4">4</div>
  <div class="boxes__box-5">5</div>
  <div class="boxes__box-6">6</div>
</div>
```


irenhh:
You did the naming right. But in your case you don't need to add separate names for bg state and order state, one general name that describe difference of the block is enough

PR #113, opened 2019-07-26
'Solution'

Serhii-P:
- https://serhii-p.github.io/layout_snake/

- https://serhii-p.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,106 @@
+body {
+  margin: 0;
+}
+
+.wrap {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+span {
+  font-size: 100px;
+  font-family: Arial;
+  color: white;
+}
+
+.block {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,106 @@
+body {
+  margin: 0;
+}
+
+.wrap {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+span {
```
irenhh: create class name for styling

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+.block {
+  height: 300px;
+  width: 33.33%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-width: 300px;
+}
+
+.one {
```
irenhh: you better specify `block-one`

PR #117, opened 2019-07-27
Snake done!

Vladyslav223:


src/style.css
```diff
@@ -0,0 +1,88 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: row wrap;
+  flex-basis: 100%;
```
irenhh: you don't need flex-basis here

PR #123, opened 2019-07-28
Snake

ibezkorovayniy:
https://ibezkorovayniy.github.io/layout_snake/
https://ibezkorovayniy.github.io/layout_snake/report/html_report/


src/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@media (min-width: 900px) {
+  .block {
+    width: 33.3%;
+  }
+
+  .container {
+    display: flex;
+    flex-direction: row;
```
irenhh: `flex-direction: row;` this is default styles for flex items

PR #124, opened 2019-07-28
layout_snake max

ArnyCorp:
- [DEMO LINK](https://arnycorporation.github.io/layout_snake/)
- [TEST REPORT LINK](https://arnycorporation.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,90 @@
+* {
+  margin: 0;
```
irenhh: if you want to reset margins of body, use it with body tag, no need to use *-selector here

src/style.css
```diff
@@ -0,0 +1,90 @@
+body {
+  margin: 0;
+  padding: 0;
```
irenhh: no need to set padding 0 for body, it doesn't have default paddings

ArnyCorp: ok, thnx!

PR #125, opened 2019-07-29
Solution

shklyarigor2015:


src/style.css
```diff
@@ -0,0 +1,98 @@
+.body {
+  margin: 0;
+  padding: 0;
```
irenhh: body doesn't have default paddings

PR #141, opened 2019-09-02
Fs aug19 andrii p

Sijey:


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.container > div {
+  font-family: Arial, sans-serif;
+  height: 300px;
+  font-size: 100px;
+  line-height: 300px;
+  text-align: center;
+  color: #fff;
+  min-width: 33.33%;
```
gestap: try to use `flex-basis` instead of `min-width`. This task for learning flex

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.block_6 {
+  background-color: #000;
+  order: 4;
+}
+
+@media (max-width: 899px) {
+  .container > div {
+    min-width: 50%;
```
gestap:
```suggestion
    flex-basis: 50%;
```

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+    order: 5;
+  }
+  .block_6 {
+    order: 6;
+  }
+}
+
+@media (max-width: 599px) {
+  .container > div {
+    min-width: 100%;
```
gestap:
```suggestion
    flex-basis: 100%;
```

PR #144, opened 2019-09-02
Solution

karpyshynroman:


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.container > div {
+  height: 300px;
+  font-family: Arial, sans-serif;
+  color: #fff;
+  font-size: 100px;
+  line-height: 300px;
+  text-align: center;
+  min-width: 33.33%;
```
gestap: try to use `flex-basis` instead of `min-width`. This task for learning flex

PR #147, opened 2019-09-02
Snake

Annetastar:


src/style.css
```diff
@@ -0,0 +1,129 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  width: 1200px;
```
gestap: this task for using flex. Try to use flex-basis

PR #233, opened 2020-02-26
completed task

Wincherz:
- [DEMO LINK](https://Wincherz.github.io/layout_snake/)
- [TEST REPORT LINK](https://Wincherz.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,67 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+}
+
+.container {
+  display: grid;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,67 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+}
+
+.container {
+  display: grid;
+  grid-template-areas: "b1 b2 b3" "b6 b5 b4";
```
earthhuman41: you need to make this task with flex, but save this solution it will come in handy in the future

PR #234, opened 2020-02-26
Solution

Aleksandr-Tyagun:
- [DEMO LINK](https://aleksandr-tyagun.github.io/layout_snake/)
    - [TEST REPORT LINK](https://aleksandr-tyagun.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,89 @@
+body {
+  margin: 0;
+  height: 100%;
+}
+
+.container {
+  display: flex;
+  justify-content: space-evenly;
+  flex-wrap: wrap;
+}
+
+.container__block {
+  font-family: Arial;
+  color: white;
+  font-size: 100px;
+  flex-grow: 1;
+  height: 300px;
+  text-align: center;
+  min-width: 300px;
+  width: 33%;
+  align-self: center;
+  line-height: 300px;
+}
+
+.container__block_1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
+}
+
+.container__block_2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.container__block_3 {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,89 @@
...
+
+.container__block_2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.container__block_3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
```
vkryvytskyy: you could not provide order value to elements that will not be change their default position. By default it will be 0, and they appear as in html order.

Aleksandr-Tyagun: Deleted unnecessary order properties in style.css

vkryvytskyy:
image
750px
image
751px

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+/*
+@media screen and (min-width:600px) and (max-width: 899px) {
+  .container__block_3 {
+    order: 4;
+  }
+  .container__block_4 {
+    order: 3;
+  }
+}
+*/
```
vkryvytskyy: remove comments

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-color: rgb(255, 0, 0);
+}
+
+.container__block_2 {
+  background-color: rgb(204, 0, 0);
+}
+
+.container__block_3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
```
vkryvytskyy: this line also could be removed

Aleksandr-Tyagun: Removed order: 3;

src/style.css
Aleksandr-Tyagun: Removed comments

PR #235, opened 2020-02-26
Solution

yurez91:
- [DEMO LINK](https://yurez91.github.io/layout_snake/)
 - [TEST REPORT LINK](https://yurez91.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  height: 300px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.block-red80 {
+  background-color: #c00;
+  order: 2;
+}
+
+.block-red60 {
+  background-color: #900;
+  order: 3;
+}
```
vkryvytskyy: you could not provide `order` value to elements, that appears in default order. By default it's set to 0, and they will display in order, that they described in html

PR #237, opened 2020-02-26
solution

hheight:
- [DEMO LINK](https://hheight.github.io/layout_snake/)
- [TEST REPORT LINK](https://hheight.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  font-family: Arial, serif;
+  color: #fff;
+  font-size: 100px;
+  height: 300px;
+  min-width: 300px;
+  width: 33%;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-grow: 1;
+}
+
+.block-1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
+}
+
+.block-2 {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.block {
+  font-family: Arial, serif;
+  color: #fff;
+  font-size: 100px;
+  height: 300px;
+  min-width: 300px;
+  width: 33%;
+
```
vkryvytskyy: remove empty line

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-grow: 1;
+}
+
+.block-1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
```

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.block-2 {
+  background-color: rgb(calc(255 / (5/4)), 0, 0);
+  order: 2;
```

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.block-3 {
+  background-color: rgb(calc(255 / (5/3)), 0, 0);
+  order: 3;
```

src/style.css
hheight: 1. it's better not to use an empty lines to visually divide, in our case, flexbox rules from others?
2. I should re-push files after fixes?

PR #238, opened 2020-02-26
Solution

ghost:
- [DEMO LINK](https://vitaliikorol.github.io/layout_snake/)
- [TEST REPORT LINK](https://vitaliikorol.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,78 @@
+body {
+  margin: 0;
+}
+
+.grid--container {
+  display: grid;
+  grid-template-areas: "grid_a_1 grid_a_2 grid_a_3"
```
earthhuman41: you need do this task with flex, save this grid solution it may come in handy in next snake task

8hokage (changes requested):

PR #239, opened 2020-02-26
solution

RustamKhananov:
[DEMO LINK](https://RustamKhananov.github.io/layout_snake/)<br>
[TEST REPORT LINK](https://RustamKhananov.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,106 @@
+
+body {
+  margin: 0;
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  flex: 1 300px;
+  height: 300px;
+  font-size: 100px;
+  text-align: center;
+  font-family: Arial, sans-serif;
+  color: white;
+  line-height: 300px ;
+}
+
+@media screen and (max-width:900px) and (min-width:600px) {
+  #block1 {
+    order: 1;
+  }
+  #block2 {
+    order: 2;
+  }
+  #block3 {
+    order: 4;
+  }
+  #block4 {
+    order: 3;
+  }
+  #block5 {
+    order: 5;
+  }
+  #block6 {
+    order: 6;
+  }
+}
+
+@media screen and (max-width: 1199px) and (min-width: 900px) {
+  #block1 {
+    order: 1;
+  }
+  #block2 {
+    order: 2;
+  }
+  #block3 {
+    order: 3;
+  }
+  #block4 {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  flex: 1 300px;
+  height: 300px;
+  font-size: 100px;
+  text-align: center;
+  font-family: Arial, sans-serif;
+  color: white;
+  line-height: 300px ;
+}
+
+@media screen and (max-width:900px) and (min-width:600px) {
```
earthhuman41: You do not need to specify max-width if you have the next breakpoint set, it will automatically be up to 900 cause then you make a request with min-width 900

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  }
+  #block5 {
+    order: 5;
+  }
+  #block6 {
+    order: 4;
+  }
+}
+
+@media screen and (min-width: 1200px) {
```
earthhuman41: You dont need to make 3 media, you can make 1 of them by default, for example this one, just write  block order directly to default styles

RustamKhananov: Default is for resolution < 600

RustamKhananov: So for 
res<900 is#1
res<1199 is#2
res = 1200 is#3 i need it because 300px*4 = 1200 but i need just 3 blocks

earthhuman41: yes, but one of them you can make default

earthhuman41: maybe try max-width, and remove 600, just left max-width 900

earthhuman41: btw you dont need spicify all blocks again in every media query, just those which order was changed

RustamKhananov: Defolt order I already use for res <600
@media screen and (min-width: 600px) for     600<res<900
@media screen and (min-width: 900px) for     900<res<1199
@media screen and (min-width: 1200px) for   1200<res
 I don`t see what I coold delete.
If I will use "max-width 900"  that will force me to make new @media for res<600

RustamKhananov: Defolt order I already use for res <600
@media screen and (min-width: 600px) for     600<res<900
@media screen and (min-width: 900px) for     900<res<1199
@media screen and (min-width: 1200px) for   1200<res
 I don`t see what I coold delete.
If I will use "max-width 900"  that will force me to make new @media for res<600

PR #241, opened 2020-02-26
Solution

KostyaPenyavskiy:
* [DEMO LINK](https://KostyaPenyavskiy.github.io/layout_snake/)
* [TEST REPORT LINK](https://KostyaPenyavskiy.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,79 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.item {
+  flex-grow: 1;
+  height: 300px;
+  width: 33%;
+  min-width: 300px;
+  margin: 0;
+  font-size: 100px;
+  line-height: 300px;
+  font-family: Arial;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  font-size: 100px;
+  line-height: 300px;
+  font-family: Arial;
+  text-align: center;
+  color: white;
+}
+
+.block_1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
```
vkryvytskyy: you don't need to provide order prop to those elements, which order should be the same as in html

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.block_2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.block_3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
```

PR #242, opened 2020-02-26
Develop

euxos:
- [DEMO LINK](https://euxos.github.io/layout_snake/)
 - [TEST REPORT LINK](https://euxos.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,75 @@
+body {
+  margin: 0;
+}
+
+.snake {
+  display: grid;
```
vkryvytskyy: make it with flex, this solution save for later, we have task anti snake, which should be maded with grid

PR #253, opened 2020-02-27
Solution v0.1

ShapovalDenys:
https://ShapovalDenys.github.io/layout_snake/

https://ShapovalDenys.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,80 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.flex {
+  height: 300px;
+  min-width: 300px;
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
+  color: white;
+  text-align: center;
+  line-height: 300px;
+  flex-grow: 1;
+  flex-basis: 33%;
+}
+
+.box_1 {
+  background-color: hsl(0,100%,50%);
+  order: 1;
+}
+
+.box_2 {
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
image
on screen 451px-600px

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+  color: white;
+  text-align: center;
+  line-height: 300px;
+  flex-grow: 1;
+  flex-basis: 33%;
+}
+
+.box_1 {
+  background-color: hsl(0,100%,50%);
+  order: 1;
```

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.box_2 {
+  background-color: hsl(0,100%,40%);
+  order: 2;
+}
+
+.box_3 {
+  background-color: hsl(0,100%,30%);
+  order: 3;
```

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.box_1 {
+  background-color: hsl(0,100%,50%);
+  order: 1;
+}
+
+.box_2 {
+  background-color: hsl(0,100%,40%);
+  order: 2;
```

src/style.css
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+@media screen and (min-width: 617px) and (max-width: 916px) {
+  .box_1 {
+    order: 1;
+  }
+  .box_2 {
+    order: 2;
+  }
```
vkryvytskyy: in media queries write only those properties, that should be changed. Omit those, that stays the same

PR #254, opened 2020-02-27
Develop

euxos:
- [DEMO LINK](https://euxos.github.io/layout_snake/)
- [TEST REPORT LINK](https://euxos.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  color: #fff;
+  font-size: 100px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.snake-part-1 {
+  background-color: hsl(0, 100%, 50%);
+  order: 1;
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.snake-part-2 {
+  background-color: hsl(0, 100%, 40%);
+  order: 2;
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.snake-part-3 {
+  background-color: hsl(0, 100%, 30%);
+  order: 3;
```

PR #256, opened 2020-02-27
solution2

ZhupanovOlexii:


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.item-2 {
+  background-color: rgb(204,0,0) ;
+
+}
+
+.item-3 {
+  background-color: rgb(153,0,0) ;
+  order: 1;
```

PR #259, opened 2020-02-27
Pls check

panasiuk-yuriy:
- [DEMO LINK](https://panasiuk-yuriy.github.io/layout_snake/)
- [TEST REPORT LINK](https://panasiuk-yuriy.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,106 @@
...
+
+  .box-5 {
+    background-color: rgb(51,0,0);
+    order: 3;
+  }
+
+  .box-6 {
+    background-color: rgb(0,0,0);
+    order: 4;
+  }
```
vkryvytskyy: in different media queries only order properties are changing. Remove bg-color from media quieries

PR #268, opened 2020-02-27
flex

VakarchukDV:
[DEMO LINK](https://VakarchukDV.github.io/layout_snake/)
 [TEST REPORT LINK](https://VakarchukDV.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,67 @@
+.wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  background-color: black;
+}
+
+.block {
+  height: 300px;
+  font-size: 100px;
+  color: #fff;
+  font-family: Arial;
+  text-align: center;
+  line-height: 300px;
+  width: 33.33%;
+  min-width: 300px;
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
image
remove body margisn

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+  text-align: center;
+  line-height: 300px;
+  width: 33.33%;
+  min-width: 300px;
+  flex-grow: 1;
+}
+
+.item_1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
```

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.item_2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
```

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.item_3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
```

PR #270, opened 2020-02-27
solution

OliinykKostya:
- [DEMO LINK](https://OliinykKostya.github.io/layout_snake/)
- [TEST REPORT LINK](https://OliinykKostya.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,75 @@
+body {
+  margin: 0;
+}
+
+.conteiner {
+  display: flex;
+  flex-wrap: wrap;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  background-color: rgb(255, 0, 0);
+}
+
+.second-block {
+  background-color: rgb(205, 0, 0);
+}
+
+.third-block {
+  background-color: rgb(154, 0, 0);
+  order: 3;
```

PR #278, opened 2020-02-28
Develop

Nastia-rosy:
- [DEMO LINK](https://Nastia-rosy.github.io/layout_snake/)
 - [TEST REPORT LINK](https://Nastia-rosy.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+
+@media screen and (max-width: 450px) {
+  .snake {
+    flex-wrap: nowrap;
+    flex-direction: column;
+  }
+
+  .items {
+    width: 100%;
+  }
```
vkryvytskyy: some of this rules is redundant

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  height: 300px;
+}
+
+.snake_head {
+  background-color: #f00;
+}
+
+.snake_tail {
+  background-color: black;
+  flex-wrap: wrap-reverse;
```
vkryvytskyy: image

src/style.css
Nastia-rosy: When I remove something from this, it stops working as it should

src/style.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+.part_3 {
+  background-color: #900;
+}
+
+.part_4 {
+  background-color: #600;
+  order: 3;
+  flex-wrap: wrap-reverse;
```
earthhuman41: why do you need this?

PR #281, opened 2020-02-28
develop

vitaliyanohin:
[DEMO LINK](https://vitaliyanohin.github.io/layout_snake/)
[TEST REPORT LINK](https://vitaliyanohin.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,81 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+}
+
+.container > div {
+  flex-basis: calc(100%/3);
+  min-width: 300px;
+  height: 300px;
+  line-height: 300px;
+  text-align: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+}
+
+.container > div {
```
vkryvytskyy: Try to avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins. Setting width and height for all DIV tags is bad practice

vkryvytskyy: Try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

PR #284, opened 2020-02-28
solution

omelchenkom:
- [DEMO LINK](https://omelchenkom.github.io/layout_snake/)
- [TEST REPORT LINK](https://omelchenkom.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,66 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  .snake-block4 {
+    order: 2;
+  }
+
+  .snake-block5,
+  .snake-block6 {
+    order: 4;
+  }
```
vkryvytskyy: same order for two different blocks?

PR #295, opened 2020-04-23
solution

aleksey-10:
- [DEMO LINK](https://aleksey-10.github.io/layout_snake/)
- [TEST REPORT LINK](https://aleksey-10.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,80 @@
+body {
+  margin: 0;
+}
+
+#snake {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,72 @@
+body {
+  margin: 0;
+}
+
+#snake {
```
irenhh, line 5: no styling by id

PR #296, opened 2020-04-23
Solution

SerhiiiOselskyi:
- [DEMO LINK](https://bran4u.github.io/layout_snake/)
- [TEST REPORT LINK](https://bran4u.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,86 @@
+body {
+  margin: 0;
+}
+
+.container {
+  max-width: 1200px;
```
irenhh, line 6: that's not necessary

irenhh (approved):

src/style.css
bran4u: I know, but somewhy cannot solve the test without it.

PR #301, opened 2020-04-23
Solution

kvalenty:
- [DEMO LINK](https://kvalenty.github.io/layout_snake/)
- [TEST REPORT LINK](https://kvalenty.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  font-size: 100px;
+  color: #fff;
+}
+
+.flex {
+  display: flex;
+}
+
+.items {
+  flex-wrap: wrap;
```
irenhh, line 13: display styles are better to write separately to each necessary element

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  max-width: 1200px;
```
irenhh, line 14: no need to limit the width

PR #331, opened 2020-07-01
solution

ArtemiiKab:
- [DEMO LINK](https://ArtemiiKab.github.io/layout_snake/)
 - [TEST REPORT LINK](https://ArtemiiKab.github.io/layout_snake/report/html_report/)


ArtemiiKab:
I think I got it. Because we have min-width 300px for div actual checks should be on 899px and 599px to prevent whitespace and order issues

PR #348, opened 2020-07-03
ready

kostyasabada:
- [DEMO LINK](https://kostyasabada.github.io/layout_snake/)
- [TEST REPORT LINK](https://kostyasabada.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,35 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,35 @@
...
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="block__section">
```
irenhh, line 14: if you try to use BEM, use it correctly
there is no block with `block` class

src/style.css
```diff
@@ -0,0 +1,78 @@
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
```
irenhh, line 4: assign size to those blocks you need

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  margin: 0;
+}
+
+.block__section {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.each__block {
```
irenhh, line 13: block_item?

src/index.html
kostyasabada: Yet didn't learn BEM so changed without BEM

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.block-5 {
+  background-color: #300;
+  order: 5;
+}
+
+.block-6 {
+  background-color: black;
+  order: 6;
+}
```
vkryvytskyy, line 50: you don't need to provide orders, if they're the same as order in html

vkryvytskyy: I mean, first block already first in html and so on

kostyasabada: If I delete 5 and 6 it doesn't work

PR #350, opened 2020-07-03
created snake

Totowa95:
- [DEMO LINK](https://Totowa95.github.io/layout_snake/)
    - [TEST REPORT LINK](https://Totowa95.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,105 @@
+html {
+  font-family: Arial, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  font-size: 100px;
+  line-height: 300px;
+  color: white;
+  flex-wrap: wrap;
+  text-align: center;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+    order: 3;
+  }
+
+  .block3 {
+    order: 4;
+  }
+
+  .block5 {
+    order: 4;
+  }
```
vkryvytskyy, line 70: two blocks with same order?

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+
+.block1 {
+  background-color: rgb(255,0,0);
+  order: 1;
+}
+
+.block2 {
+  background-color: rgb(204,0,0);
+  order: 2;
+}
```
vkryvytskyy, line 33: this blocks always are first and second, no need to provide orders for them

PR #359, opened 2020-07-05
snake was born

YuriiRysyn:
- [DEMO LINK](https://YuriiRysyn.github.io/layout_snake/)
- [TEST REPORT LINK](https://YuriiRysyn.github.io/layout_snake/report/html_report/)

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,0 +1,1 @@
+[0705/161001.467:ERROR:file_io.cc(89)] ReadExactly: expected 36, observed 0
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,94 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  color: #fff;
+  font-family: arial, sans-serif;
+  font-size: 100px;
+}
+
+.box {
+  display: flex;
+  height: 300px;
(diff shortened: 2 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  .fifth_container {
+    order: 5;
+  }
+
+  .fourth_container {
+    order: 6;
+  }
+}
+
+@media screen and (min-width:600px) and (max-width:899px) {
```
vkryvytskyy, line 64: you can provide only min/max width, needed rules will apply when you reach new screen size

vkryvytskyy:
зображення
What's that?

debug.log
```diff
@@ -0,0 +1 @@
+[0705/161001.467:ERROR:file_io.cc(89)] ReadExactly: expected 36, observed 0
```
danheim, line 1: You shouldn't add this file to the pull request.

src/style.css
```diff
@@ -0,0 +1,104 @@
...
+  flex-wrap: wrap;
+  color: #fff;
+  font-family: arial, sans-serif;
+  font-size: 100px;
+}
+
+.box {
+  display: flex;
+  height: 300px;
+  min-width: 300px;
```
danheim, line 16: use flex basis

PR #363, opened 2020-07-05
HTML & CSS added

Chelovekozavr:
[LINK](https://Chelovekozavr.github.io/layout_snake/)
[TEST](https://Chelovekozavr.github.io/layout_snake/report/html_report/))

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+}
+
+.conteiner {
+  display: flex;
+  flex-wrap: wrap;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.block2 {
+  order: 2;
+  background-color: #c00;
+}
+
+.block3 {
+  order: 3;
+  background-color: #900;
+}
```
vkryvytskyy, line 34: no need to provide order to elements that appear in the same order in html

vkryvytskyy:
зображення
Block width should never be less than 300px

Chelovekozavr:
Got it. fixed

PR #380, opened 2020-07-06
Solution

martakupka:
- [DEMO LINK](https://martakupka.github.io/layout_snake/)
- [TEST REPORT LINK](https://martakupka.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,109 @@
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
+}
+
+.block-container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  text-align: center;
+  line-height: 300px;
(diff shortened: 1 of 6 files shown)


vkryvytskyy:
зображення

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+    flex-grow: 1;
+  }
+
+  .block-1 {
+    order: 1;
+  }
+
+  .block-2 {
+    order: 2;
+  }
```
vkryvytskyy, line 62: this blocks are always first and second

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  .block-6 {
+    order: 6;
+  }
+}
+
+@media screen and (min-width: 900px) {
+  .block {
+    flex-grow: 1;
+  }
+
```
vkryvytskyy, line 85: this should be default property

PR #382, opened 2020-07-06
flex

RusTron:
- [DEMO LINK](https://rustron.github.io/layout_snake/)
- [TEST REPORT LINK](https://rustron.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  flex: 0 0 100%;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+
+.number5 {
+  background-color: #300;
+  order: 5;
+}
+
+.number6 {
+  background-color: #000;
+  order: 6;
+}
```
vsm2k17, line 48: They already in this order no need to write order explicitly

src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  .number6 {
+    order: 4;
+  }
+
+  .number4 {
+    order: 6;
+  }
+}
+
+@media (min-width: 600px) and (max-width: 899px) {
```
vsm2k17, line 64: no need to use max-width and min-width, you can remove one of them

PR #412, opened 2020-08-28
solution

KhrystynaYelyseyeva:
- [DEMO LINK](https://KhrystynaYelyseyeva.github.io/layout_snake/)
- [TEST REPORT LINK](https://KhrystynaYelyseyeva.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,84 @@
+*,
+*::after,
+*::before {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: white;
+  text-align: center;
+  line-height: 300px;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.flex__container {
+  background: black;
+
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: stretch;
+  align-content: stretch;
+}
+
+.flex__item {
+  height: 300px;
+  min-width: 300px;
+
+  flex: 33.3%;
+}
+
+.flex__item__one {
+  background: rgba(255, 0, 0, 1);
+}
+
+.flex__item__two {
+  background: rgba(255, 0, 0, 0.8);
+}
+
+.flex__item__three {
+  background: rgba(255, 0, 0, 0.6);
+  order: 3;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+*::before {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: white;
+  text-align: center;
+  line-height: 300px;
```
YuriiHoliuk, line 12: do not use line-height to set block height

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
YuriiHoliuk, line 17: body doesn't have a default padding

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.flex__container {
+  background: black;
+
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
```
YuriiHoliuk, line 25: flex-flow

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.flex__item {
+  height: 300px;
+  min-width: 300px;
+
+  flex: 33.3%;
+}
+
+.flex__item__one {
```
YuriiHoliuk, line 38: there is no element of the element in BEM

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.flex__item__four {
+  background: rgba(255, 0, 0, 0.4);
+  order: 4;
+}
+
+.flex__item__five {
+  background: rgba(255, 0, 0, 0.2);
+  order: 5;
```
YuriiHoliuk, line 58: it is default one

YuriiHoliuk (changes requested):
Good job! I like your code!

src/style.css
KhrystynaYelyseyeva: I use this to align the text vertically

src/style.css
KhrystynaYelyseyeva: Sorry, but the default value is zero

YuriiHoliuk (approved):

PR #421, opened 2020-08-30
snake sssh

ssergeta:
[DEMO LINK](https://ssergeta.github.io/layout_snake/)
[TEST REPORT LINK](https://ssergeta.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,95 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+  font-size: 100px;
+}
+
+.blocks {
+  display: flex;
+  flex-wrap: wrap;
+  color: white;
+}
+
+.block {
+  flex-basis: 33.33%;
+  height: 300px;
+  width: 300px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.one {
+  background-color: #f00;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,95 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+  font-size: 100px;
```
YuriiHoliuk, line 8: better to set font styles in html

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.six {
+  background-color: #000;
+}
+
+@media (max-width: 899px) {
+  .block {
+    flex-basis: 50%;
+  }
```
YuriiHoliuk, line 53: try to achieve the same effect without changing flex-basis

YuriiHoliuk (changes requested):
Good job!

src/style.css
ssergeta: ready, add it to " * "

YuriiHoliuk: - comment was to font-family and font-size
- HTML does not mean *

ssergeta: > * comment was to font-family and font-size
> * HTML does not mean *

ready

YuriiHoliuk (approved):
🔥🤘🏻

PR #423, opened 2020-08-30
Completed Snake

obashmakov:
- [DEMO LINK](https://obashmakov.github.io/layout_snake/)
- [TEST REPORT LINK](https://obashmakov.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,97 @@
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  height: 300px;
+  min-width: 300px;
+
+  text-align: center;
+  line-height: 300px;
+  color: #fff;
+  font-size: 100px;
+}
+
+.box_1 {
+  background-color: #f00;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,97 @@
+body {
+  font-family: Arial, Helvetica, sans-serif;
```
YuriiHoliuk, line 2: better to set in html

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.box_6 {
+  background-color: #000;
+}
+
+@media screen and (min-width: 900px) {
+  .box {
+    flex-basis: 33.33%;
+  }
```
YuriiHoliuk, line 48: try to achieve the same effect without changing flex-basis

YuriiHoliuk (changes requested):

src/style.css
obashmakov: Ok, fixed that!

src/style.css
obashmakov: Here I understand from slack that don't need to change.

YuriiHoliuk (approved):

PR #470, opened 2020-11-13
Added 6 flex-blocks

olesiatk:
- [DEMO LINK](https://olesiatk.github.io/layout_snake/)
- [TEST REPORT LINK](https://olesiatk.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,86 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.block {
+  height: 300px;
+  width: 300px;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
+  text-align: center;
+  line-height: 300px;
+  flex: 1 0 33.3%;
+}
+
+.block-1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
+}
+
+.block-2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.block-3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
+
+.block-4 {
+  background-color: rgb(102, 0, 0);
+  order: 6;
+}
+
+.block-5 {
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+  order: 4;
+}
+
+@media screen and (max-width: 899px) {
+  .container div {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+
+.block-2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.block-3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
```
vsm2k17, line 39: they had these values by default no need to write them explicitly

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  order: 5;
+}
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+  order: 4;
+}
+
+@media screen and (max-width: 899px) {
+  .container div {
```
vsm2k17, line 57:
```suggestion
  .container .block {
```

vsm2k17 (changes requested):
delete redundant

src/style.css
olesiatk: changed to class

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.block {
+  height: 300px;
+  width: 300px;
+  font-family: Arial, Helvetica, sans-serif;
```
aionova, line 18: indicate font-family in html or body tag

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+
+.block-4 {
+  background-color: rgb(102, 0, 0);
+  order: 6;
+}
+
+.block-5 {
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
```
aionova, line 46: order: 5 is default for block-5

src/style.css
```diff
@@ -0,0 +1,83 @@
...
+    order: 4;
+  }
+
+  .block-4 {
+    order: 3;
+  }
+
+  .block-6 {
+    order: 6;
+  }
```
aionova, line 68: same. order:6 is default

aionova (changes requested):
Еще немного правок и будет супер)

src/style.css
olesiatk: If I take away this value the order of elements changes

src/style.css
olesiatk: The first value for .block-6 is `order: 4;`
Then it changes to `order: 6;`

aionova: sorry, my fault!

src/style.css
olesiatk: ok, changed

PR #474, opened 2020-11-13
Snake solution

Roman-Matsuk:
- [DEMO LINK](https://Roman-Matsuk.github.io/layout_snake/)
- [TEST REPORT LINK](https://Roman-Matsuk.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,81 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  height: 300px;
+  width: 300px;
+  flex-basis: 300px;
+  color: #fff;
+  justify-content: center;
+  align-items: center;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  flex-grow: 1;
+}
+
+.default-first {
+  background-color: rgb(255, 0, 0);
+  order: 1;
+}
+
+.default-second {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.default-third {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  justify-content: center;
+  align-items: center;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  flex-grow: 1;
+}
+
+.default-first {
+  background-color: rgb(255, 0, 0);
+  order: 1;
```
vsm2k17, line 25: it's already 1 by default, no need to write this explicitly

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
+
+.default-last {
+  background-color: rgb(0, 0, 0);
+  order: 4;
+}
+
+@media screen and (max-width: 750px) {
```
vsm2k17, line 53: 750? why this point?

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  .default-fourth {
+    order: 3;
+  }
+
+  .default-last {
+    order: 6;
+  }
+}
+
+@media screen and (max-width: 450px) {
```
vsm2k17, line 67: 450?

vsm2k17 (changes requested):
image
something wrong don't you think?

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+
+.default-second {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.default-third {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
```
vsm2k17, line 35: These elements also had the same order by default

vsm2k17 (approved):
good work

PR #480, opened 2020-11-14
'snake' was created

HaidaiAndrii:
- [DEMO LINK](https://HaidaiAndrii.github.io/layout_snake/)
- [TEST REPORT LINK](https://HaidaiAndrii.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,97 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: white;
+}
+
+.divs {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.blocks {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 300px;
+  width: 33.33%;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,97 @@
...
+
+.block1 {
+  background-color: red;
+  order: 1;
+}
+
+.block2 {
+  order: 2;
+  background-color: #c00;
+}
```
aionova, line 30: order 1 for block1 is already set by default. No need to indicate it

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+    width: 50%;
+  }
+
+  .block1 {
+    order: 1;
+  }
+
+  .block2 {
+    order: 2;
+  }
```
aionova, line 64: No need to repeat the values, that you set previously.

aionova (changes requested):
Немного подправить и будет огонь!)

aionova (approved):
Отлично!

PR #483, opened 2020-11-14
flex-box

DarMatt:
- [DEMO LINK](https://DarMatt.github.io/layout_snake/)
- [TEST REPORT LINK](https://DarMatt.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,107 @@
+* {
+  box-sizing: border-box;
+}
+
+.html {
+  font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  font-size: 100px;
+  color: white;
+  justify-content: center;
+  align-items: center;
+}
+
+.block {
+  height: 300px;
+  min-width: 300px;
+  flex: 0 0 33.3%;
+  text-align: center;
+  line-height: 300px;
+  font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
+}
+
+.one {
+  background: red;
+  order: 1;
+}
+
+.two {
+  background: hsl(0, 100%, 40%);
+  order: 2;
+}
+
+.three {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,107 @@
...
+  align-items: center;
+}
+
+.block {
+  height: 300px;
+  min-width: 300px;
+  flex: 0 0 33.3%;
+  text-align: center;
+  line-height: 300px;
+  font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
```
aionova, line 28: no need to repeat font-family. You inherit it by default

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+}
+
+.one {
+  background: red;
+  order: 1;
```
aionova, line 33: order: 1 is set by default for '.one'. No need to set it

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+}
+
+.two {
+  background: hsl(0, 100%, 40%);
+  order: 2;
```
aionova, line 38: same

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+.six {
+  background: hsl(0, 0%, 0%);
+  order: 4;
+}
+
+@media screen and (max-width: 899px) {
+  .block {
+    flex-basis: 50%;
+  }
+  .one {
```
aionova, line 65: same. Indicate only rules, that you want to change. Don't list all of them

aionova (changes requested):

src/style.css
DarMatt: @aionova does not inherit, i checked

DarMatt: When i delete font-family from .block the style is not applied

aionova (changes requested):

src/style.css
DarMatt: @aionova ooooooo thanks))

aionova (approved):
Good job! :rocket:

PR #573, opened 2021-03-29
layout-snake

dmytro-abram:
[DEMO LINK](https://dmytro-abram.github.io/layout_snake/)
[TEST REPORT LINK](https://dmytro-abram.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,81 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+body {
+  margin: 0;
+}
+
+.conteiner {
+  display: flex;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-color: #300;
+  order: 5;
+}
+
+.box6 {
+  background-color: #000;
+  order: 3;
+}
+
+@media (max-width: 750px) {
```
vsm2k17, line 55: the width of your blocks must be at least 300 pixels, if you use a media query similar to yours, this does not satisfy the condition

vsm2k17: image

vsm2k17 (changes requested):

PR #583, opened 2021-03-29
solution

IvanKolodiazhnyi:
- [DEMO LINK](https://IvanKolodiazhnyi.github.io/layout_snake/)
- [TEST REPORT LINK](https://IvanKolodiazhnyi.github.io/layout_snake/report/html_report/)


vsm2k17 (changes requested):
let's add a condition about block behaviour on screens bigger than 900px, each block width should be 33.3%
Screenshot 2021-03-31 at 12 36 16

PR #585, opened 2021-03-30
Snake

starzhitskiy:
- [DEMO LINK](https://starzhitskiy.github.io/layout_snake/)
- [TEST REPORT LINK](https://starzhitskiy.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
-</html>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  .three {
+    order: 1;
+  }
+}
+
+.four {
+  background-color: rgb(102, 0, 0);
+}
+
+@media (min-width: 600px) {
```
varseniuk, line 64: Don't duplicate your media queries. You may write once 
`@media (min-width: 600px) { }`
and write all relevant to this width styles there (not write query many times)

varseniuk (changes requested):
Your solution is working, but let's make it using `flex-basis`.
Also, you should write all styles, relevant to the specific media-query in one selector, not in many identical

src/index.html
```diff
@@ -1,14 +1,22 @@
...
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
-</html>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+        >
```
varseniuk, line 7:
```suggestion
    <meta
      name="viewport"
      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
    >
```

varseniuk (approved):
Good job! 👍 
But, probably, there are some problems with your deploy, because your demo link looks weird (though the code is good and everything works fine)



Redeploy pls

PR #590, opened 2021-03-30
solution

den-bulaev:
- [DEMO LINK](https://den-bulaev.github.io/layout_snake/)
- [TEST REPORT LINK](https://den-bulaev.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,103 @@
+body {
+  margin: 0;
+}
+
+.parent-block {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+}
+
+.child-block {
+  display: flex;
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,103 @@
...
+  height: 300px;
+  min-width: 300px;
+  color: white;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+}
+
+@media (min-width: 900px) {
+  .parent-block {
+    height: 600px;
```
vsm2k17, line 24: interesting solution 👍, but usually if you work with the width you would use flex-basis.

vsm2k17 (approved):
good job

PR #592, opened 2021-03-30
snake solution_01

ivan-shpynda:
- [DEMO LINK](https://ivan-shpynda.github.io/layout_snake/)
- [TEST REPORT LINK](https://ivan-shpynda.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -4,100 +4,88 @@ body {
   padding: 0;
 }
 
-.block-1 {
-  background-color: #f00;
-  flex-grow: 1;
-}
-
-.block-2 {
-  background-color: #c00;
-  flex-grow: 1;
-}
-
-.block-3 {
-  background-color: #900;
-  flex-grow: 1;
-}
-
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,103 @@
+body {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+.block-1 {
+  background-color: #f00;
+  flex-grow: 1;
```
varseniuk, line 9: Don't repeat yourself. You may write `flex-grow: 1;` in your `.block` class and it will be applied to all blocks

varseniuk (changes requested):
Let's use flex-basis for this task

varseniuk (approved):
Your code is working, but as I've said earlier - it would be better to use `flex-basis` property here and not stick to static widths.
Please, pay more attention to my comments in future 👍

PR #594, opened 2021-03-30
Snake task

leonid-vegera:
- [DEMO LINK](https://leonid-vegera.github.io/layout_snake/)
- [TEST REPORT LINK](https://leonid-vegera.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,84 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+}
+
+body {
+  margin: 0;
+  --min-box-height: 300px;
+  --max-row-width: 1199px;
+}
+
+.row {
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  --min-box-height: 300px;
+  --max-row-width: 1199px;
+}
+
+.row {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  max-width: var(--max-row-width);
```
varseniuk, line 21:
```suggestion
```

Let's not use max-width for this task and use 'flex-basis' instead. You result looks not like it should be:



Two rows should be stretched to full width of the page

varseniuk (changes requested):
Let's use 'flex-basis' and not 'max-width'

varseniuk (approved):
Great! 👍

PR #597, opened 2021-03-30
solution

Allrightman:
[DEMO LINK](https://Allrightman.github.io/layout_snake/)
-[TEST REPORT LINK](https://Allrightman.github.io/layout_snake/report/html_report/)


vsm2k17 (changes requested):
almost done, let's add a condition, that on screens bigger than 900 px each block width would be 33.3% of the screen

vsm2k17 (approved):
great

PR #603, opened 2021-03-30
solution

Namart:
[DEMO LINK](https://Namart.github.io/layout_snake/)
[TEST REPORT LINK](https://Namart.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
varseniuk, line 6:
```suggestion
    <meta
      name="viewport"
      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
    >
```

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@media(min-width: 300px) {
+  .container {
+    flex-direction: column;
+  }
+  .box {
+    flex-basis: 100%;
+  }
+}
```
varseniuk, line 51:
```suggestion
```

You can easily remove this media-query because you have `min-width: 600px` and `min-width: 900px` queries, so in any other cases, which differ from those two screen widths (that means <600px) will be used styles written above.
And no need to change flex-direction to column, because you use `flex-wrap:wrap` and all your blocks will be moving to the next line if they not fit the screen width

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+
+@media (min-width:600px) {
+  .container {
+    flex-direction: row;
+  }
```
varseniuk, line 56:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  .box-6 {
+    order: 3;
+  }
+}
+
+@media (min-width:900px) {
+  .container {
+    flex-direction: row;
+  }
```
varseniuk, line 77:
```suggestion
```

varseniuk (changes requested):
Great job!
Read my comment and let's remove one redundant media-query

PR #606, opened 2021-03-31
The «Snake» task was done

GoldenSpade:
### Snake	
- [DEMO LINK](https://BilliZane.github.io/layout_snake/)
- [TEST REPORT LINK](https://billizane.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/.bemlintrc.json
+++ b/.bemlintrc.json
@@ -1,0 +1,16 @@
+{
+  "elementDivider": "__",
+  "modifierDivider": "_",
+  "ignore": [
+    "node_modules",
+    "dist"
+  ],
+  "rules": {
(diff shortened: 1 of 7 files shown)


varseniuk (changes requested):
Let's use `flex-basis` instead of writing manually `width` for every media query.
In fact, you need only two breakpoints - 600px and 900px. According to this screen widths, you should split your flex-container into two (50%) or 3 (33%) columns.

Also, there should not be maximum width. On wide screens, all your blocks should stretch into two rows and three columns. Now your task looks wrong:

varseniuk (approved):
👍

PR #607, opened 2021-03-31
layout_snake solution

Grygoriy-Shytikov:
- [DEMO LINK](https://Grygoriy-Shytikov.github.io/layout_snake/)
- [TEST REPORT LINK](https://Grygoriy-Shytikov.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,113 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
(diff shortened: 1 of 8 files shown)


src/style.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.block_6 {
+  background-color: hsl(0, 100%, 0%);
+}
+
+@media (min-width: 300px) {
+  .container {
+    flex-direction: column;
+  }
```
varseniuk, line 52:
```suggestion
```
Just FYI - you may not change flex-direction at all (here and in other media queries) just by giving your `.block` property `flex-basis: 100%` (like you do in your `min-width: 900px` query)

varseniuk (approved):
Good! But please pay attention my comment and take it into consideration for future

PR #608, opened 2021-03-31
snake project

RuslanaPodshyvalova:
- [DEMO LINK](https://RuslanaPodshyvalova.github.io/layout_snake/)
- [TEST REPORT LINK](https://RuslanaPodshyvalova.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,6 +12,17 @@
   </head>
   <body>
     <div class="container">
+      <div class="box box1">1</div>
+      <div class="box box2">2</div>
+      <div class="box box3">3</div>
+      <div class="box box4">4</div>
+      <div class="box box5">5</div>
+      <div class="box box6">6</div>
+    </div>
+
+
+    <!--
+    <div class="container">
       <div class="column">
         <div class="group__box1">
           <div class="box box1">1</div>
@@ -43,5 +54,7 @@
         </div>
       </div>
     </div>
+
+  -->
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,47 @@
...
+    <div class="container">
+      <div class="column">
+        <div class="group__box1">
+          <div class="box box1">1</div>
+          <div class="box box2">2</div>
+          <div class="box box3">3</div>
+          <div class="box box4">4</div>
+          <div class="box box5">5</div>
+          <div class="box box6">6</div>
+        </div>
```
varseniuk, line 23: It should be only one group of boxes and it should change according to the screen size.
Use media-queries for this.
For screen with a width up to 300px the blocks should be set up from top to bottom (as in your `group__box3`)
for width from 301px to 899px it should be two columns and three rows (as in your `group__box2`)
and for a width 900px and more it should be two rows with three columns (as in your `group_box1`).
All these blocks should be adaptive and change in size according to the screen width

varseniuk (changes requested):
Probably, you have misunderstood the task requirements. There shoul be one container with six blocks and they should resize and change their position adaptively, according to the screen width.
You've made three different containers, but there's should be only one and position of blocks must be as follow:

varseniuk (approved):
Just FYI - you may also use `flex-basis` property and give the proper percentages for different media queries.
Also - remove commented lines in your code

PR #612, opened 2021-03-31
develop

Dima-Batiuk:
- [DEMO LINK](https://Dima-Batiuk.github.io/layout_snake/)
- [TEST REPORT LINK](https://Dima-Batiuk.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,111 @@
+html,
+body,
+div {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.blocks {
+  display: flex;
+  height: 300px;
+  min-width: 300px;
+  flex-grow: 1;
+  color: #fff;
+  justify-content: center;
+  align-items: center;
+}
+
+.block_1 {
+  background-color: rgb(255, 0, 0);
+}
+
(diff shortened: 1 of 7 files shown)


vsm2k17 (changes requested):
image
should on full screen

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+body,
+div {
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+  margin: 0;
+  padding: 0;
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
+}
```
vsm2k17, line 11: a lot of those properties is redundant for body and div tags, for example, fonts are inherited, if you write them once for html tag it would be enough

src/style.css
```diff
@@ -0,0 +1,113 @@
...
+
+@media (min-width: 600px) {
+
+  .blocks {
+    flex-basis: 50%;
+  }
+
+  .block_1 {
+    order: -6;
```
vsm2k17, line 68: why -6, better to use positive values

vsm2k17 (changes requested):
very good, just couple comments

src/style.css
Dima-Batiuk: It was easier for me in that way)

vsm2k17 (approved):
great

vsm2k17 (changes requested):
!test

vsm2k17 (approved):
!test

PR #698, opened 2021-05-31
first try to pass the snake task

ErnestMk1:
DEMO LINK - https://ErnestMk1.github.io/layout_snake/
TEST REPORT LINK - https://ErnestMk1.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,94 @@
+* {
+  margin: 0;
+}
+
+.general {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,94 @@
+* {
+  margin: 0;
+}
```
aionova, line 3: don't use * selector for margins. Use it only for box-sizing property

aionova (approved):
Good job!

PR #699, opened 2021-05-31
install commit

NazarYanchyshyn:
- [DEMO LINK](https://NazarYanchyshyn.github.io/layout_snake/)
- [TEST REPORT LINK](https://NazarYanchyshyn.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,80 @@
+* {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  flex-grow: 1;
+}
+
+.flex_block {
+  flex-basis: 33%;
+  height: 300px;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,80 @@
+* {
+  margin: 0;
```
aionova, line 2: don't set margin: 0 for all the elements. Use * selector only for box-sizing

aionova (changes requested):
Minimal width of each block should be 300px

aionova (approved):
Great!

aionova (approved):

YuriiHoliuk (commented):
ff

YuriiHoliuk (approved):

PR #708, opened 2021-06-03
Task done

Peephie:
https://peephie.github.io/layout_snake/
https://peephie.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,23 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,85 @@
...
+
+.block-1 {
+  background-color: #f00;
+  order: 1;
+}
+
+.block-2 {
+  background-color: #c00;
+  order: 2;
+}
```
vkryvytskyy, line 41: this two blocks always are first and second

src/index.html
```diff
@@ -1,14 +1,23 @@
...
+  <body>
+    <!-- <h1>Snake</h1> -->
+    <div class="blocks">
+      <div class="block block-1">1</div>
+      <div class="block block-2">2</div>
+      <div class="block block-3">3</div>
+      <div class="break-row"></div> <!--I used this div to divide blocks into at least two rows-->
+      <div class="block block-4">4</div>
+      <div class="block block-5">5</div>
+      <div class="block block-6">6</div>
```
vkryvytskyy, line 20:
```suggestion
      <div class="block">1</div>
      <div class="block">2</div>
      <div class="block">3</div>
      <div class="block">4</div>
      <div class="block">5</div>
      <div class="block">6</div>
```
use nth-child to apply background styles based on element's position in DOM}
Also you don't need no any other elements to divide blocks

vkryvytskyy (dismissed):
use flex-basis to define how much space each block should use

src/index.html
Peephie: What do I need to use to make the layout at least 2 rows? By the way i already use flex-basis with value 300px in .block.

vpolets (approved):

PR #740, opened 2021-07-07
Snake V1.0

ihor-shapoval:
- [DEMO LINK](https://ihor-shapoval.github.io/layout_snake/)
- [TEST REPORT LINK](https://ihor-shapoval.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,68 @@
+* {
+  margin: 0;
+}
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+.block {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.snake__block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 300px;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  min-width: 300px;
+  flex-basis: 33%;
+  flex-grow: 1;
+}
+
+.snake__block:first-child {
+  background-color: #f00;
```
varseniuk, line 27: It's not a bad solution with :nth-child but let's create a separate class for each block, e.g block-1, block-2 etc. These classes should be responsible for color and order.
And use a mix of classes like `class="block block-1"
Also, pay attention to BEM - when you are using class `snake__block` it means that it is the element of the block `snake`. But you don't have such block

varseniuk (changes requested):
Your logic is correct, but let's use mix of classes instead of nth-child

varseniuk (approved):

PR #745, opened 2021-07-07
snake was created

dima-tkachenko:
- [DEMO LINK](https://dima-tkachenko.github.io/layout_snake/)
- [TEST REPORT LINK](https://dima-tkachenko.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,107 @@
+body {
+  margin: 0;
+  padding: 0;
+  color: #fff;
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.columns {
+  display: flex;
+  flex-wrap: wrap;
+}
+  
+.block {
+  display: flex;
+  height: 300px;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,107 @@
...
+  height: 300px;
+  justify-content: center;
+  align-items: center;
+  flex-basis: 100%;
+  flex-shrink: 0;
+  flex-grow: 1;
+}
+
+.block:nth-child(1) {
+  background-color: rgb(100%, 0%, 0%);
```
varseniuk, line 25: It's not a bad solution with :nth-child but let's create a separate class for each block, e.g block-1, block-2 etc. These classes should be responsible for color and order.
And use a mix of classes like `class="block block-1"`

varseniuk (changes requested):
Your logic is correct, but let's use a mix of classes instead of nth-child

varseniuk (approved):

PR #749, opened 2021-07-08
i did this task

RomSmile:
- DEMO LINK https://RomSmile.github.io/layout_snake/
- TEST REPORT LINK https://RomSmile.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+html {
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.main_container {
+  display: flex;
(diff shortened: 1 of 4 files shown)


RomSmile:
http://joxi.ru/4Ak66Vqi0MNxV2

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.box6 {
+  background-color: #000;
+  order: 1;
+}
+
+@media (max-width: 899px) {
+  .box {
+    width: 50%;
```
varseniuk, line 57: You could use `flex-basis` instead of `width`

varseniuk (approved):
Logic is good. But as we are studying flexbox. it would be great to use `flex-basis` instead of `width`.

PR #752, opened 2021-07-08
task is done

Oleksandr-Zhitenko:


src/style.css
```diff
@@ -0,0 +1,65 @@
+body {
+  display: flex;
```
varseniuk, line 2: Don't use `display:flex` for whole body.
Make some container-wrapper and give display: flex to it

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  font-family: "Arial", sans-serif;
+  font-size: 100px;
+  color: #fff;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.box:nth-child(2) {
+  background-color: #c00;
```
varseniuk, line 22: You have classes like `box1`, `box2` etc. Why are you using `nth-child` instead of them?
Let's change it

varseniuk (changes requested):
Make flex container for your boxes and use class names instead of `nth-child`

src/style.css
Oleksandr-Zhitenko: why using display:flex for body is mistake?

Ihor-Karpyn: Because if you need to add another block to the body, flex applies to it too

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  background-color: #300;
+  order: 2;
+}
+
+.box6 {
+  background-color: #000;
+  order: 1;
+}
+
+@media only screen and (max-width: 750px) {
```
Ihor-Karpyn, line 45:
```suggestion
@media only screen and (max-width: 899px) {
```

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  .box:nth-child(1n) {
+    order: 0;
+  }
+
+  .box:nth-child(1n+3):not(.box4) {
+    order: 1;
+  }
+}
+
+@media only screen and (max-width: 450px) {
```
Ihor-Karpyn, line 58:
```suggestion
@media only screen and (max-width: 599px) {
```

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  .box {
+    flex-basis: 50%;
+  }
+  .box:nth-child(1n) {
+    order: 0;
+  }
+
+  .box:nth-child(1n+3):not(.box4) {
+    order: 1;
+  }
```
Ihor-Karpyn, line 55: You have a class like box1, don't use nth:child, it is difficult understanding

Ihor-Karpyn (changes requested):
Almost done 🔥, need some changes.

PR #753, opened 2021-07-08
Snake 

alex-glotov67:
- [DEMO LINK](https://alex-glotov67.github.io/layout_snake/)
- [TEST REPORT LINK](https://alex-glotov67.github.io/layout_snake/report/html_report/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "0.0.1",
-    "@mate-academy/scripts": "^0.9.1",
+    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/stylelint-config": "*",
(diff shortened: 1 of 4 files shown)


ihor-jpeg:
The problem with your code is that when the width becomes larger than 450px, the second box drops back into the first row. The it turns out that you have 2 boxes for the total width of 450px, but the assignment specifically asks you to make sure that the width of each box is not smaller than its height, which is 300px. The point is, you need to make sure that there are two boxes in a row only then the page width is equal or bigger than 600px.

alex-glotov67:
> The problem with your code is that when the page width becomes larger than 450px, the second box drops back into the first row. Then it turns out that you have 2 boxes for the total width of 450px (225px per box), but the assignment specifically asks you to make sure that the width of each box is not smaller than its height, which is 300px. The point is, you need to make sure that there are two boxes in a row only when the page width is equal to or bigger than 600px.


Thanks a lot! My mistake. Already fix it!

danheim (approved):

PR #757, opened 2021-07-08
done

AndreySerhovetc:
- [DEMO LINK](https://AndreySerhovetc.github.io/layout_snake/)
- [TEST REPORT LINK](https://AndreySerhovetc.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,88 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  align-items: center;
+  justify-content: center;
+  width: 33.3%;
+  min-width: 300px;
+  height: 300px;
+  font-size: 100px;
+  color: #fff;
+}
+
+.container > :nth-child(1) {
```
varseniuk, line 27:
```suggestion
.container > :nth-child(1) {
```
It's not a bad soulution with `:nth-child` but let's create separate class for each block, e.g `box-1`, `box-2` etc. These classes should be responsible for colour and order.
 And use a mix of classes like `class="box box-1"

varseniuk (changes requested):
Your logic is correct, but let's use mix of classes instead of `nth-child`

varseniuk (approved):

PR #783, opened 2021-07-10
Done snake

Uran-web:
- [DEMO LINK](https://Uran-web.github.io/layout_snake/)
- [TEST REPORT LINK](https://Uran-web.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,106 @@
+* {
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  width: 100%;
+}
+
+.box {
+  height: 300px;
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
+  color: #fff;
+
+  flex-basis: 300px;
+  flex-grow: 1;
+  flex-shrink: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+}
+
+.box-1 {
+  background-color: rgb(255, 0, 0);
+}
+
+.box-2 {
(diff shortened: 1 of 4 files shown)


VinnieJ-2k20 (changes requested):
Nice start.
Just fix some details and we're OK!

You still have default `body` margin. That causes tests to fail.
Also, minimum amount of rows should be 2, you have 1 here:
image

On 1300px some strange thing happens: 
image

On 620px order changes, but number of rows - don't
image

VinnieJ-2k20 (changes requested):
You have too many breakpoints.
Number of rows and order should change only on 600 and 900 px marks.
Check the screenshots again.

After 1200px blocks order is reset again, but it should be the same as on example.
image

VinnieJ-2k20 (changes requested):
image

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+.box-6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media screen {
+  .container {
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+}
```
VinnieJ-2k20, line 59: This can be moved to general `.container` selector.

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  .box-6 {
+    order: 2;
+  }
+  .box-5 {
+    order: 3;
+  }
+  .box-4 {
+    order: 4;
+  }
+}
```
VinnieJ-2k20, line 74: If you use `desktop-first` approach this can also be used without media query.

VinnieJ-2k20 (approved):
Good job 👍

PR #788, opened 2021-07-11
snake done

yevhenpecherskyi:
https://yevhenpecherskyi.github.io/layout_snake/
https://yevhenpecherskyi.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="row">
```
VinnieJ-2k20, line 12: Let's use more appropriate class name.
It's not a `row`. To me, it looks more like `container`.

src/index.html
```diff
@@ -1,14 +1,21 @@
...
+      <div class="container num1">1</div>
```
VinnieJ-2k20, line 13: And this is not a container. Let's call it `block` or `box`, what you prefer more.

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+.num6 {
+  background-color: black;
+}
+
+@media all {
+  .row {
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+}
```
VinnieJ-2k20, line 54: `all` is default value, just use it under regular selector.

VinnieJ-2k20 (changes requested):
Almost done.
Need to fix couple of things.

After 1200px blocks should still occupy full page width:
image

Also, added some comments on code.

VinnieJ-2k20 (approved):
Good job 👍

PR #791, opened 2021-07-12
snake

00-22-11:
[DEMO LINK](https://00-22-11.github.io/layout_snake/)
[TEST REPORT LINK](https://00-22-11.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,81 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.block--5 {
+  background-color: #300;
+  order: 5;
+}
+
+.block--6 {
+  background-color: #000;
+  align-self: inherit;
```
VinnieJ-2k20, line 54: Not sure we need it here. Control using only order.

VinnieJ-2k20 (changes requested):
Some strange things happening with order in breakpoints:

600 px:
image

900 px:
image

Also, after 900px some white line appears on the right, therefore the test fail:
image

VinnieJ-2k20 (approved):

PR #834, opened 2021-08-19
snake solution

Daerniss:
- [DEMO LINK](https://daerniss.github.io/layout_snake/)
- [TEST REPORT LINK](https://daerniss.github.io/layout_snake/report/html_report/)


Ihor-Karpyn (approved):
The class will better than nth-child

PR #897, opened 2021-09-10
sssssnake

arielg0rl:
[DEMO LINK](https://arielg0rl.github.io/layout_snake/)
[TEST REPORT LINK](https://arielg0rl.github.io/layout_snake/report/html_report/)


VinnieJ-2k20 (changes requested):
Solution looks almost good. There should not be extra horizontal space on the right on wide screens
image

VinnieJ-2k20 (approved):

PR #932, opened 2021-10-12
okok

ghost:
- [DEMO LINK](https://junglq.github.io/layout_snake/)
- [TEST REPORT LINK](https://junglq.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,5 +1,8 @@
-body {
+* {
   margin: 0;
+}
+
+body {
   font-family: Arial, sans-serif;
 }
 
@@ -9,8 +12,7 @@ body {
 }
 
 .wrapper__block {
-  height: 300px;
-  width: 310px;
+  min-height: 300px;
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+.wrapper__block {
+  height: 300px;
+  width: 310px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  font-size: 100px;
+  color: #fff;
+  flex-basis: 33.33%;
```
AlinaMatuschak, line 19: Can use calc

AlinaMatuschak (changes requested):
Good start 👌. But check small details

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.wrapper__block {
+  min-height: 300px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  font-size: 100px;
+  color: #fff;
+  flex-basis: 33.33%;
```
AlinaMatuschak, line 21: Is the idea with calc not interesting?

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.wrapper__block:nth-child(5) {
+  background-color: rgb(20%, 0%, 0%);
+  order: 5;
+}
+
+.wrapper__block:nth-child(6) {
+  background-color: rgb(0%, 0%, 0%);
+  order: 4;
```
AlinaMatuschak, line 49: you can star from 1

AlinaMatuschak (changes requested):

src/style.css
junglq: dont understand

src/style.css
AlinaMatuschak: Why?

src/style.css
AlinaMatuschak: You don't have 1, 2, 3 orders. It's not a mistake, but try not to use too big values )

src/style.css
AlinaMatuschak: It helps to read your code easily. Try to avoid fractional values

PR #958, opened 2021-10-16
first try

PavloKhanas:
[Demo link](https://pavlokhanas.github.io/layout_snake/)
[Test result link](https://pavlokhanas.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -2,28 +2,26 @@
   box-sizing: border-box;
 }
 
-body {
+html {
   font-family: Arial, sans-serif;
-  margin: 0;
-  padding: 0;
 }
 
-div {
-  display: block;
+body {
+  margin: 0;
 }
 
 .blocks {
   display: flex;
   flex-wrap: wrap;
-  font-size: 100px;
-  text-align: center;
-  color: white;
-  line-height: 300px;
 }
 
 .block {
   height: 300px;
-  flex-basis: 33.3%;
+  flex-basis: calc(100% / 3);
+  font-size: 100px;
+  text-align: center;
+  color: white;
+  line-height: 300px;
 }
 
 .block_1 {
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Arial, sans-serif;
```
illiarerun, line 6: It should be in html selector

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  margin: 0;
+  padding: 0;
```
illiarerun, line 8: Body doesn't have default paddings

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+div {
+  display: block;
+}
```
illiarerun, line 13: Unnecessary, divs have display: block by default
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+.blocks {
+  display: flex;
+  flex-wrap: wrap;
+  font-size: 100px;
+  text-align: center;
+  color: white;
+  line-height: 300px;
```
illiarerun, line 21: Move these properties to .block class, it's not right to set style properties of children to wrapper class

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.block {
+  height: 300px;
+  flex-basis: 33.3%;
```
illiarerun, line 26: Let's try with calc

ihor-jpeg (approved):
Well done! there is one suggestion regarding the order of the flex elements. you had to write the order prop for every display size. but when you think about it, the default order is the perfect order for the max-width: 599px media query. so my suggestion would be to position the boxes with "mobile first" approach. you don't set the order prop initially. then you change the order of the 3 and 4 for tablets, and then on full screen you change the order of 4, 5, and 6. it is not critical, but in my opinion the code will be more structured and easier to understand.

PR #959, opened 2021-10-16
layout snake

polina-latun:


src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  margin: 0;
+}
+
+@media (min-width: 900px) {
+  .blockContainer {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: stretch;
+    min-width: 900px;
+  }
```
illiarerun, line 11: Also there is no sense to repeat code in every media query, move your .blockContainer class out of them, and remove min-width

PR #973, opened 2021-10-20
Snake

WorkAccCom:
- [DEMO LINK](https://workacccom.github.io/layout_snake/)
- [TEST REPORT LINK](https://workacccom.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,93 @@
+@import url(./styles/reset.css);
+
+.html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.blocks {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.blocks__block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  min-width: 300px;
+  font-size: 100px;
+  color: white;
+  flex-grow: 1;
+  flex-basis: 33.33%;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  height: 300px;
+  min-width: 300px;
+  font-size: 100px;
+  color: white;
+  flex-grow: 1;
+  flex-basis: 33.33%;
+}
+
+.blocks__block--1 {
+  background-color: rgb(100% 0% 0%);
```
aionova, line 25: in rgb you need to indicate not percent, but a number. that's why deployed version doesn't show  colors

aionova (changes requested):

src/style.css
WorkAccCom: For me, it's very strange, because in my local project it works correctly.

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@media (max-width: 599px) {
+  .blocks__block--3 {
+    background-color: rgb(153, 0, 0);
+    order: 3;
+  }
+
+  .blocks__block--4 {
+    background-color: rgb(102, 0, 0);
```
aionova, line 80: you could use variables for colors

aionova (approved):

PR #1075, opened 2021-12-23
first

KostyaKovbel:
[[demo]](https://kostyakovbel.github.io/layout_snake/)
[[test]](https://kostyakovbel.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,90 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+  font-size: 100px;
+}
+
+.container {
+  display: flex;
+  margin: 0;
+  padding: 0;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.container {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  flex-direction: column;
+  justify-content: center;
+  flex-wrap: wrap;
+  width: 100%;
```
Anna-Tkachenko, line 18: container has 100 width by default

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.block {
+  height: 300px;
+  min-width: 300px;
```
Anna-Tkachenko, line 23: use flex-basis for setting width

Anna-Tkachenko (approved):

PR #1083, opened 2021-12-23
successful

AndryKK:


src/index.html
```diff
@@ -1,14 +1,21 @@
...
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="space">
+      <div class="Block B1">1</div>
+      <div class="Block B2">2</div>
+      <div class="Block B3">3</div>
+      <div class="Block B4">4</div>
+      <div class="Block B5">5</div>
+      <div class="Block B6">6</div>
```
Anna-Tkachenko, line 18:
```suggestion
      <div class="block block--1">1</div>
      <div class="Block B2">2</div>
      <div class="Block B3">3</div>
      <div class="Block B4">4</div>
      <div class="Block B5">5</div>
      <div class="Block B6">6</div>
```

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+body {
+  margin: 0;
+}
+
+.space {
+  display: flex;
+  flex-wrap: wrap;
+  max-width: 1199px;
```
Anna-Tkachenko, line 17: Don't set max-width, it should be always 100%

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1112, opened 2022-01-05
Develop

yahyafullstack01:
- [DEMO LINK](https://yahyafullstack01.github.io/layout_snake/)
- [TEST REPORT LINK](https://yahyafullstack01.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,5 +1,5 @@
 body {
-  background-color: aqua;
+  background-color: #ffff;
   margin: 0;
 }
 
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,109 @@
+body {
+  background-color: aqua;
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: row;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.box {
+  height: 300px;
+  min-width: 33.33%;
+  text-align: center;
+  font-size: 100px;
+  line-height: 300px;
+  color: #ffff;
+  font-family: Arial, serif;
+}
+
+.box:nth-child(1) {
+  background: rgb(255, 0, 0);
+  order: 1;
+  flex-grow: 1;
+}
+
+.box:nth-child(2) {
+  background: rgb(204, 0, 0);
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,109 @@
...
+.box:nth-child(5) {
+  background: rgb(51, 0, 0);
+  order: 5;
+  flex-grow: 1;
+}
+
+.box:nth-child(6) {
+  background: rgb(0, 0, 0);
+  order: 4;
+  flex-grow: 1;
```
vkryvytskyy, line 56: this property is similar to all boxes, move it to `.box`, so you don't need to write it for each box

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+  .box:nth-child(4) {
+    order: 4;
+  }
+
+  .box:nth-child(3) {
+    order: 3;
+  }
+
+}
```
vkryvytskyy, line 109:
```css
media (max-width: 599px) {
  // styles
}

media (max-width: 900px) {
  // styles
}
```
place your code like this, and remove min-width from first media query

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  background: rgb(255, 0, 0);
+  order: 1;
+  flex-grow: 1;
+}
+
+.box:nth-child(2) {
+  background: rgb(204, 0, 0);
+  order: 2;
+  flex-grow: 1;
+}
```
vkryvytskyy, line 33: You can omit order property for the first two boxes

src/style.css
```diff
@@ -0,0 +1,109 @@
+body {
+  background-color: #ffff;
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: row;
```
vkryvytskyy, line 8: it's default flex-flow value, you can omit it

vkryvytskyy: image
BTW, there should be wrap also, so you don't need 9 line

vkryvytskyy (changes requested):

vkryvytskyy (approved):

PR #1135, opened 2022-01-20
Create a Star

GoDfreE25:


src/style.css
```diff
@@ -0,0 +1,103 @@
...
+.blockSix {
+  background-color: #000;
+}
+
+@media (min-width: 300px) {
+
+  .block {
+    flex-basis: 100%;
+  }
+}
```
solaryasha, line 54: You don't need this media query

solaryasha: Write this rule in `block` selector

solaryasha (approved):

PR #1181, opened 2022-02-15
Snake implemented

VETALEK:
- [DEMO LINK](https://vetalek.github.io/layout_snake/)
- [TEST REPORT LINK](https://vetalek.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,4 @@
-* {
+html {
   font-family: Arial, Helvetica, sans-serif;
 }
 
@@ -7,6 +7,12 @@ body {
 }
 
 .container {
+  position: absolute;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
+
   display: flex;
   flex-wrap: wrap;
 }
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
+* {
```
illiarerun, line 1:
```suggestion
html {
```

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.block {
+  font-size: 100px;
+  color: #fff;
+  display: flex;
+  text-align: center;
+  justify-content: center;
+  align-items: center;
+  width: 300px;
```
illiarerun, line 21: use flex-basis instead

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  height: 300px;
+  flex-grow: 1;
+  flex-shrink: 0;
+}
+
+.block-1 {
+  order: 1;
```
illiarerun, line 28: that's default order, don't set it explicitly, fix it for all classes

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  .block-4 {
+    order: 3;
+  }
+}
+
+@media (min-width: 900px) {
+  .container {
+    max-width: 1199px;
```
illiarerun, line 69: don't restrict container width, it should be 3 columns on all resolutions > 1200 and also it should be fullscreen, not like that:
image

illiarerun (changes requested):
Nice start, now lets work on fixes:

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+body {
+  margin: 0;
+}
+
+.container {
+  position: absolute;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
```
illiarerun, line 14: redundant
```suggestion
```

illiarerun (approved):

PR #1206, opened 2022-02-17
Add layout_snake

YegorKochetkov:
https://YegorKochetkov.github.io/layout_snake/
https://YegorKochetkov.github.io/layout_snake/report/html_report/


src/style.css
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-grow: 1;
+
+  height: 300px;
+  min-width: 300px;
```
ivan-kadykalo, line 17: Use `flex-basis` instead of `min-width`

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  .block--6 {
+    order: 4;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    width: 33.33%;
```
ivan-kadykalo, line 72: will be better to use `calc(100% / 3)` and `flex-basis`

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+  background-color: rgb(51, 0, 0);
+}
+
+.block--6 {
+  background-color: #000;
+}
+
+@media (min-width: 600px) {
+  .block {
+    width: 50%;
```
ivan-kadykalo, line 50: here also

ivan-kadykalo (changes requested):
Nice code))
Only couple things left

ivan-kadykalo:
For some reasons your links doesn't work. 
image

MaksimusCRS (approved):
GJ!

PR #1211, opened 2022-02-18
some message

AntonShaga:
https://antonshaga.github.io/layout_snake/


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  background-color: rgb(102, 0, 0);
+}
+
+.block--5 {
+  order: 5;
+  background-color: rgb(51, 0, 0);
+}
+
+.block--6 {
+
```
danylo-onishchuk, line 46: зачем тут пустая строка?

danylo-onishchuk (approved):

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+.block--5 {
+  order: 5;
+  background-color: rgb(51, 0, 0);
+}
+
+.block--6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media (max-width: 750px) {
```
jstmpelowycz, line 48: Why 750px?

src/style.css
```diff
@@ -0,0 +1,73 @@
...
+  }
+  .block--5 {
+    order: 5;
+  }
+  .block--6 {
+    order: 6;
+  }
+}
+
+@media (max-width: 450px) {
```
jstmpelowycz, line 66: Why 450px?

jstmpelowycz: And here is the same way with `min-width`

jstmpelowycz (changes requested):
You're almost done, just apply some fixes, and you're good to go further!

PR #1256, opened 2022-03-17
Done

Ksenia-Didukh:
- [DEMO LINK](https://Ksenia-Didukh.github.io/layout_snake/)
- [TEST REPORT LINK](https://Ksenia-Didukh.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,220 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 300px;
+  width: 300px;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+  color: #fff;
+}
+
+.block-1 {
+  background-color: #f00;
+}
+
+.block-2 {
+  background-color: rgb(204 0 0);
+}
+
+.block-3 {
+  background-color: rgb(153, 0, 0);
+}
+
+.block-4 {
+  background-color: rgb(102, 0, 0);
+}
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,220 @@
...
+}
+
+@media (min-width: 600px) {
+  .block {
+    width: 300px;
+  }
+
+  .blocks {
+    display: flex;
+    flex-wrap: wrap;
```
Bohdan-Maksymiuk, line 54: you should use it be default on all widths

src/style.css
```diff
@@ -0,0 +1,220 @@
...
+    order: 3;
+  }
+  
+  .block-5 {
+    order: 5;
+  }
+  
+  .block-6 {
+    order: 6;
+  }
```
Bohdan-Maksymiuk, line 79: no need to change order for for all blocks

src/style.css
```diff
@@ -0,0 +1,220 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 300px;
+  width: 300px;
```
Bohdan-Maksymiuk, line 7:
```suggestion
  min-width: 300px;
```

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-03-18 at 13 03 26
Your snake should always be stretched to full-screen width.
There could be only 2 media. Use min-width and mb other flex properties (flex-grow, flex-basis)

Bohdan-Maksymiuk (approved):
Great job 👍

PR #1258, opened 2022-03-18
index.html and styles are created, tests are passed.

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/layout_snake/)
- [TEST REPORT LINK](https://efdalyalcin.github.io/layout_snake/report/html_report/)

I have used; 
@media (min-width:900px) {
  .boxes { min-width:30vw; }
}

I guess there can be a better way but I couldn't understand why it doesn't work with 33vw but work with 30vw?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,80 @@
+body {
+  margin: 0;
+
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+}
+
+.container {
+  display: flex;
+  flex-flow: wrap;
```
VinnieJ-2k20, line 10: Don't use shorthand property, if you don't benefit from it.
To set one rule - just use `flex-wrap`.

src/index.html
```diff
@@ -1,14 +1,21 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="container">
+      <div class="boxes box1">1</div>
```
VinnieJ-2k20, line 13: It's one box, so let's use appropriate name
```suggestion
      <div class="box box1">1</div>
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the coments.

PR #1259, opened 2022-03-18
added html layout, styles and updated readme

LukaszGryziewicz:
- [DEMO LINK](https://LukaszGryziewicz.github.io/layout_snake/)
- [TEST REPORT LINK](https://LukaszGryziewicz.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
+}
+
+.box--1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
+}
+
+.box--2 {
+  background-color: rgb(204, 0, 0);
+  order: 2;
+}
+
+.box--3 {
+  background-color: rgb(153, 0, 0);
+  order: 3;
+}
+
+.box--4 {
+  background-color: rgb(102, 0, 0);
+  order: 4;
+}
+
+.box--5 {
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
+
+.box--6 {
+  background: rgb(0, 0, 0);
+  order: 6;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
```
ihor-jpeg, line 22: can you achieve the same without `flex-grow`?

ihor-jpeg: you really only need `flex-basis`
less is more

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.box--5 {
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
+
+.box--6 {
+  background: rgb(0, 0, 0);
+  order: 6;
+}
```
ihor-jpeg, line 53: this seems to be the default behavior in terms of order

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+  .box--4 {
+    order: 3;
+  }
+}
+
+@media (min-width: 900px) {
+  .box {
+    flex-basis: calc(100% / 3);
+  }
```
ihor-jpeg, line 68: 🥇

ihor-jpeg (changes requested):
Very close! Nice one!

src/style.css
LukaszGryziewicz: Entire_document_750. & Entire_document_450  fail without flex-grow

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
```
ihor-jpeg, line 22: well, it's not just simply removing `flex-grow`. you have hardcoded `flex-basis` to 300px. it will fail like this. but you can use percents

ihor-jpeg (changes requested):
One tiny thing left!

src/style.css
ihor-jpeg: think about the behavior that you want and how you can achieve it

ihor-jpeg (approved):
Well done!

PR #1260, opened 2022-03-18
some message

rekhakilari:
- [DEMO LINK](https://rekhakilari.github.io/layout_snake/)
- [TEST REPORT LINK](https://rekhakilari.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
         content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
-<body>
-  <h1>Snake</h1>
-</body>
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="container">
+      <div class="block1 block">1</div>
```
VinnieJ-2k20, line 13: Recommended order of classes - general ones first, then specific.
```suggestion
      <div class="block block1 ">1</div>
```

src/style.css
```diff
@@ -0,0 +1,88 @@
+body {
+  padding: 0;
```
VinnieJ-2k20, line 2: `body` has no default paddings, so this rule is redundant.

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  .block5 {
+    order: 2;
+  }
+
+  .block6 {
+    order: 1;
+  }
+}
```
VinnieJ-2k20, line 63: You can use this as default styles without media query.

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+
+@media (min-width: 600px) and (max-width: 899px) {
```
VinnieJ-2k20, line 65: If you use code above as default, you won't need the first condition.
```suggestion
@media (max-width: 899px) {
```

VinnieJ-2k20 (approved):
Good job 👍 

Check the comments for some advice for future tasks,

PR #1263, opened 2022-03-19
snake task html & css updated

amitpatiljc:
- [DEMO LINK](https://amitpatiljc.github.io/layout_snake/)
- [TEST REPORT LINK](https://amitpatiljc.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,103 @@
+body {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: white;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  justify-content: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  .block-3 {
+    order: 4;
+  }
+
+  .block-4 {
+    order: 3;
+  }
+
+  .block-6 {
+    background-color: black;
```
VinnieJ-2k20, line 66: This block doesn't change color depending on screen width.
So, the background-color rule here is redundant.

VinnieJ-2k20 (approved):
Good job 👍 
Added one small comment.

src/style.css
amitpatiljc: Yes I think this i forgot to delete after copy pasting from code above default .block-6. I will delete it.
thank you

PR #1264, opened 2022-03-19
src & readme

elkhayate:
- [DEMO LINK](https://elkhayate.github.io/layout_snake/)
- [TEST REPORT LINK](https://elkhayate.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,45 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="wrap">
+      <div class="red block-one">
+        <span>
+        1
+        </span>
+      </div>
+      <div class="red block-two">
+        <span>
+        2
+        </span>
+      </div>
+      <div class="red block-three">
+        <span>
+        3
+        </span>
+      </div>
+      <div class="red block-six">
+        <span>
+        6
+        </span>
+      </div>
+      <div class="red block-five">
+        <span>
+        5
+        </span>
+      </div>
+      <div class="red block-four">
+        <span>
+        4
+        </span>
+      </div>
+    </div>
+  </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,106 @@
+body {
+  margin: 0;
+}
+
+.wrap {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+span {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
+}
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,45 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="wrap">
+      <div class="red block-one">
```
VinnieJ-2k20, line 13: Don't limit yourself with class names.
What if this block would have different color.
Better to use more general classnames.
```suggestion
      <div class="block block-one">
```

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+.wrap {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+span {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
+}
```
VinnieJ-2k20, line 14: Don't style using tags, use classes.

src/index.html
```diff
@@ -1,14 +1,45 @@
...
+      <div class="red block-five">
+        <span>
+        5
+        </span>
+      </div>
+      <div class="red block-four">
+        <span>
+        4
+        </span>
+      </div>
```
VinnieJ-2k20, line 42: Work with the natural order of the elements - 1, 2, 3, 4, 5, 6.

src/style.css
```diff
@@ -0,0 +1,109 @@
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
```
ihor-jpeg, line 5: let's use hex

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  flex-wrap: wrap;
+}
+
+.block {
+  height: 300px;
+  width: 33.33%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-width: 300px;
```
ihor-jpeg, line 19: but why?

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+}
+
+.wrap {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  height: 300px;
+  width: 33.33%;
```
ihor-jpeg, line 15: this is a flex container. let's learn flex and use `flex-basis`. also 33.33% is not quite what we want, because 33.33 * 3 is 99.99, but not 100. let's use `calc(100% / 3)`

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+.block-two {
+  background-color: #c00;
+  order: 2;
+}
+
+.block-three {
+  background-color: #900;
+  order: 3;
+}
```
ihor-jpeg, line 35: it seems to me that blocks 1-3 will naturally have this order

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+.block-six {
+  background-color: #000;
+  order: 4;
+}
+
+@media (min-width: 600px) and (max-width: 899px) {
+  .block {
+    width: 50%;
+  }
```
ihor-jpeg, line 55: let's use flex-basis

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+
+  .block-six {
+    order: 6;
+  }
+}
+
+@media (max-width: 599px) {
+  .block {
+    width: 100%;
+  }
```
ihor-jpeg, line 85: flex-basis

ihor-jpeg (changes requested):
Almost there!

Regarding the media queries. No need to use complex media queries, like `@media (min-width: 600px) and (max-width: 899px)`.
Simply implement a default behavior, let's say mobile. then you only need
`(min-width: 600px)` and `(min-width: 900px)`

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #1265, opened 2022-03-19
Develop

MabuzaM:
- [DEMO LINK](https://MabuzaM.github.io/layout_snake/)
- [TEST REPORT LINK](https://MabuzaM.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,105 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: row;
+  flex-wrap: nowrap;
+}
+
+.block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+  height: 300px;
+  width: 300px;
+  color: white;
+  background-color: red;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  padding: 0;
+  margin: 0;
+}
+
+.block-2 {
+  background-color: color-mix(in sRGB, red 80%, black 20%);
+}
+
+.block-3 {
+  background-color: color-mix(in sRGB, red 60%, black 40%);
+}
+
+.block-4 {
+  background-color: color-mix(in sRGB, red 40%, black 60%);
+}
+
+.block-5 {
+  background-color: color-mix(in sRGB, red 20%, black 80%);
+}
+
+.block-6 {
+  background-color: black;
+}
+
+@media (min-width: 300px),
+  (max-width: 450px),
+  (max-width: 600px),
+  (max-width: 750px),
+  (max-width: 1050px),
+  (max-width: 1200px) {
+  .container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+    flex-flow: row;
+    flex-wrap: wrap;
+  }
+}
+
+@media (max-width: 450px) {
+  .block-1,
+  .block-2,
+  .block-3,
+  .block-4,
+  .block-5,
+  .block-6 {
+    flex-grow: 1;
+  }
+}
+
+@media (min-width: 600px),
+  (max-width: 750px) {
+  .block-3 {
+    order: 4;
+  }
+
+  .block-4 {
+    order: 3;
+  }
+
+  .block-5 {
+    order: 5;
+  }
+
+  .block-6 {
+    order: 6;
+  }
+}
+
+@media (min-width: 900px),
+  (min-width: 1201px) {
+  .block-4 {
+    order: 6;
+  }
+
+  .block-5 {
+    order: 5;
+  }
+
+  .block-6 {
+    order: 4;
+  }
+}
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Boilerplate for layout tasks
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_snake/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_snake/report/html_report/)
+- [DEMO LINK](https://MabuzaM.github.io/layout_snake/)
+- [TEST REPORT LINK](https://MabuzaM.github.io/layout_snake/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline)
 
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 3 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,125 @@
...
+
+.block-5 {
+  background-color: rgb(51, 0, 0);
+}
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media (min-width: 450px),
```
VinnieJ-2k20, line 52: See checklist item 2.

You should have no more than 2 media queries.

MabuzaM: Okay thanx have removed them now

MabuzaM: BACKSTOP ERROR: Reference file not found C:\Users\User\mate_projetcs\layout_snake\backstop_data\bitmaps_reference\Entire_document_450.png

I get this error when I run the test. The test images show but the reference images don't

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+
+.block-5 {
+  background-color: rgb(51, 0, 0);
+}
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media (min-width: 450px) {
```
ihor-jpeg, line 53: why 450px?

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  .container {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
```
ihor-jpeg, line 58: you have set these styles already
DRY - don't repeat yourself

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+    order: 6;
+  }
+}
+
+@media (min-width: 900px) {
+  .container {
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+  }
```
ihor-jpeg, line 82: DRY

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+
+  .block {
+    flex-basis: 33.33%;
+  }
```
ihor-jpeg, line 86: 33.33*3 is not quite 100

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  flex-grow: 1;
+  height: 300px;
+  width: 300px;
+  color: white;
+  background-color: rgb(255, 0, 0);
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  padding: 0;
+  margin: 0;
+}
```
ihor-jpeg, line 31: whoa, i think that's excessive

ihor-jpeg (changes requested):
Almost there! But your links do not work. Make sure you have deployed your project

src/style.css
MabuzaM: Noted. Have tried to fix

src/style.css
MabuzaM: Yes it's 99.99999999999...... Am trying to find a way to do it :)

src/style.css
MabuzaM: Tried to optimize

src/style.css
MabuzaM: Yes noted thank you

src/style.css
MabuzaM: Saw it in the mockup. Have used 600px

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  .block-6 {
+    order: 6;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: 33.33%;
```
ihor-jpeg, line 74: try `calc(100% / 3)`

ihor-jpeg (changes requested):
Almost there!

src/style.css
MabuzaM: Okay thank you. tried to divide without calc(). So to calculate in CSS, do we always use the calc() function or it's only in this case?

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  justify-content: center;
+  align-items: center;
+  padding: 0;
+  margin: 0;
+}
+
+.block {
+  flex-grow: 1;
+  height: 300px;
+  width: 300px;
```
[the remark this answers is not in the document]
MabuzaM: Okay noted

src/style.css
ihor-jpeg: if you need 50%, then no need to use calc. 50% is exactly the same as calc(100% / 2). calc is performed by the browser and if you have too many of them, the performance might drop.
but in this case, there is no way to set 33.3333.....% without calc, so we should use it

ihor-jpeg: @MabuzaM

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.container {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  padding: 0;
+  margin: 0;
```
ihor-jpeg, line 22: your container is a `div`. `divs` do not have any default margins or paddings

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+* {
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
```
ihor-jpeg, line 20: I think these also do not do anything. your content inside of the container takes up the entire space. nothing to center

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+    order: 3;
+  }
+
+  .block-5 {
+    order: 5;
+  }
+
+  .block-6 {
+    order: 6;
+  }
```
ihor-jpeg, line 73:
```suggestion
  .block-5,
  .block-6 {
    order: 5;
  }
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  background-color: rgb(51, 0, 0);
+}
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media (min-width: 600px) {
+  .block {
+    flex-basis: calc(100% / 2);
```
ihor-jpeg, line 57: here you don't really need calc

ihor-jpeg (changes requested):
Well done! Let's clean up the styles

src/style.css
MabuzaM: Noted thank you

src/style.css
MabuzaM: Okay changed. Thank you

src/style.css
MabuzaM: Noted thanx. fixed

src/style.css
MabuzaM: Okay noted thank you

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+  flex-direction: row;
```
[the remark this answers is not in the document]
MabuzaM: Noted thank you

src/style.css
MabuzaM: Thank you @ihor-jpeg understood

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  flex-grow: 1;
```
ihor-jpeg, line 21: flex frow here seems not needed

ihor-jpeg (approved):
Well done!

PR #1266, opened 2022-03-20
The desing based on snake have been created using html and css

leny-jeanbaptiste:
[DEMO LINK](https://leny-jeanbaptiste.github.io/layout_snake/)
[TEST REPORT LINK](https://leny-jeanbaptiste.github.io/layout_snake/report/html_report/)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="container">
+      <div class="block b1">1</div>
+      <div class="block b2">2</div>
+      <div class="block b3">3</div>
+      <div class="block b4">4</div>
+      <div class="block b5">5</div>
+      <div class="block b6">6</div>
```
ihor-jpeg, line 18: no no no. `b1` is not a good class name. when creating class names use the whole word
`block--1` is much better

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+
+html {
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: rgb(255, 255, 255);
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
ihor-jpeg, line 13: body does not have any default paddings

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  flex-basis: 100%;
+  flex-grow: 1;
```
ihor-jpeg, line 27: after you have set `flex-basis`, `flex-grow` does not do anything

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+
+.b5 {
+  background-color: rgb(51,0,0);
+  order: 5;
+}
+
+.b6 {
+  background-color: rgb(0,0,0);
+  order: 6;
+}
```
ihor-jpeg, line 58: the order here is redundant. this is the default behavior. try to learn writing less code

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+
+  .b5 {
+    background-color: rgb(51,0,0);
+    order: 5;
+  }
+
+  .b6 {
+    background-color: rgb(0,0,0);
+    order: 4;
+  }
```
ihor-jpeg, line 100: you seem to be repeating the same color. the only thing that needs to be changed is the order

src/style.css
```diff
@@ -0,0 +1,102 @@
...
+  .b5 {
+    order: 5;
+  }
+
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: 33%;
+  }
```
ihor-jpeg, line 82: 33 + 33 + 33 is 99, not 100
that's why when this media query is applied you get this
Screenshot 2022-03-21 at 22 31 41

the white line on the right, because it does not cover 100%. only 99

ihor-jpeg (changes requested):
Well done! Let's finish it!

src/style.css
ihor-jpeg: the only thing it does here is it saves you from the issue in my last comment. but it can be avoided if you set the `flex-basis` properly there

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: calc(100% / 3);
+  }
+
+  .block--3 {
+    order: 3;
+  }
```
ihor-jpeg, line 83: what about this one?

src/style.css
ihor-jpeg: do not `flex-grow` anymore

ihor-jpeg (changes requested):
Almost there!

ihor-jpeg (approved):
Nicely done! 🚀

PR #1267, opened 2022-03-21
snake layout

iLebo-stack:
https://iLebo-stack.github.io/layout_snake/
https://iLebo-stack.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,26 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width,
+          user-scalable=no,
+          initial-scale=1.0,
+          maximum-scale=1.0,
+          minimum-scale=1.0"
+    >
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="blocks">
+      <div class="block block--1">1</div>
+      <div class="block block--2">2</div>
+      <div class="block block--3">3</div>
+      <div class="block block--4">4</div>
+      <div class="block block--5">5</div>
+      <div class="block block--6">6</div>
+    </div>
+  </body>
 </html>
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Boilerplate for layout tasks
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_snake/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_snake/report/html_report/)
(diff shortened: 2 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,26 @@
...
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width,
+          user-scalable=no,
+          initial-scale=1.0,
+          maximum-scale=1.0,
+          minimum-scale=1.0"
+    >
```
ihor-jpeg, line 11: think about indentation. `content` is the property, after that, we have this prop's values. they should be indented more than the `content`

```suggestion
    <meta name="viewport"
          content="
            width=device-width,
            user-scalable=no,
            initial-scale=1.0,
            maximum-scale=1.0,
            minimum-scale=1.0
          "
    >
```

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  margin: 0;
+}
+
+.block {
+  height: 300px;
+  line-height: 300px;
+  text-align: center;
+  font-family: Arial, sans-serif;
+  color: #fff;
+  font-size: 100px;
```
ihor-jpeg, line 11: it's okay. but generally speaking, you want to do that in the html tag

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.block--1 {
+  background-color: #f00;
+}
+
+.block--2 {
+  background-color: rgb(204, 0, 0);
+}
```
ihor-jpeg, line 20: be consistent. use either hex or rgb

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  background-color: rgb(102, 0, 0);
+}
+
+.block--5 {
+  background-color: rgb(51, 0, 0);
+}
+
+.block--6 {
+  background-color: #000;
+}
```
ihor-jpeg, line 36: same here

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+
+@media (min-width: 600px) {
+  .blocks {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  .block {
+    flex-basis: calc(100 / 2 * 1%);
+  }
```
ihor-jpeg, line 46:
```suggestion
  .block {
    flex-basis: calc(100% / 2);
  }
```

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+
+  .block--1 {
+    order: 1;
+  }
+  .block--2 {
+    order: 2;
+  }
```
ihor-jpeg, line 53: this is the defualt behavior

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  .block--3 {
+    order: 3;
+  }
```
ihor-jpeg, line 81: default behavior

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  .block--6 {
+    order: 6;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: calc(100 / 3 * 1%);
+  }
```
ihor-jpeg, line 71:
```suggestion
  .block {
    flex-basis: calc(100% / 3);
  }
```

ihor-jpeg (changes requested):
Almost there!

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+@media (min-width: 600px) {
+  .blocks {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  .block {
+    flex-basis: calc(100% / 2);
```
ihor-jpeg, line 48: if you have many calc functions it might get heavy for the browser. here you can simply use 50%

ihor-jpeg (approved):
Well done! Take a look at the comment

PR #1268, opened 2022-03-21
Snake task

huberciak:
- [DEMO LINK](https://huberciak.github.io/layout_snake/)
- [TEST REPORT LINK](https://huberciak.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,93 @@
+html {
+  color: white;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  font-size: 100px;
+}
+
+.block {
+  display: flex;
+  line-height: 300px;
+  justify-content: center;
+  height: 300px;
+  min-height: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
+}
+
+.block--1 {
+  background-color: rgba(255,0,0,255);
+}
+
+.block--2 {
+  background-color: rgba(204,0,0,255);
+  order: 1;
+}
+
+.block--3 {
+  background-color: rgba(153,0,0,255);
+  order: 2;
+}
+
+.block--4 {
+  background-color: rgba(102,0,0,255);
+  order: 3;
+}
+
+.block--5 {
+  background-color: rgba(51,0,0,255);
+  order: 4;
+}
+
+.block--6 {
+  background-color: rgba(0,0,0,255);
+  order: 5;
+}
+
+@media (min-width: 600px) {
+  .block--3 {
+    order: 4;
+  }
+
+  .block--4 {
+    order: 3;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,93 @@
+html {
+  color: white;
```
ihor-jpeg, line 2: let's use hex for colors. 
or rgb

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  font-size: 100px;
```
ihor-jpeg, line 13: font size can be set in html. keep things together

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+.block {
+  display: flex;
+  line-height: 300px;
+  justify-content: center;
+  height: 300px;
+  min-height: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
```
ihor-jpeg, line 23: why use `min-height` here?

you can achieve the same without `flex-grow` if you do not hardcode `flex-basis` in pixels. think about other units

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+
+.block--5 {
+  background-color: rgba(51,0,0,255);
+  order: 4;
+}
+
+.block--6 {
+  background-color: rgba(0,0,0,255);
+  order: 5;
+}
```
ihor-jpeg, line 53: this order seems strange. and i don't think you actually need it

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+
+  .block--4 {
+    order: 3;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: 33%;
+  }
```
ihor-jpeg, line 68: 33% * 3 is 99%, not 100%. the reason it works is that you have `flex-grow` above.  thin how you can get 33.3333333.......%

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  .block--4 {
+    order: 5;
+  }
+
+  .block--5 {
+    order: 4;
+  }
+
+  .block--6 {
+    order: 3;
```
ihor-jpeg, line 91: if you look at the result, you will find that the first three elements are located in their places even without the order
write less code

ihor-jpeg (changes requested):
Good work! Let's fix some errors

src/style.css
```diff
@@ -0,0 +1,86 @@
...
+    order: 1;
+  }
+
+  .block--5 {
+    order: 3;
+  }
+
+  .block--6 {
+    order: 3;
+  }
```
ihor-jpeg, line 67: if two or more selectors have identical styles it's a good idea to combine them

```suggestion
  .block--5,
  .block--6 {
    order: 3;
  }
```

PR #1269, opened 2022-03-22
Initial commit - WiP

rHashTag:
- [DEMO LINK](https://rHashTag.github.io/layout_snake/)
- [TEST REPORT LINK](https://rHashTag.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,68 @@
+body {
+  padding: 0;
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  max-width: 900px;
+
+}
+
+@media (min-width: 300px) {
+  .container {
+    height: 1800px;
+  }
+}
+
+@media (min-width: 600px) {
+  .container {
+    height: 900px;
+  }
+}
+
+@media (min-width: 900px) {
+  .container {
+    height: 600px;
+  }
+}
+
+.block {
+  font-family: Arial, sans-serif;
+  color: rgb(255, 255, 255);
+  font-size: 100px;
+
+  flex-basis: 300px;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  font-family: Arial, sans-serif;
+  color: rgb(255, 255, 255);
+  font-size: 100px;
```
ihor-jpeg, line 14: it's alright. but generally, you want to set it for the HTML tag and only change it for selectors that use a different font size or color. font-family, for the most part, stays the same across the website

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+  width: 100%;
```
ihor-jpeg, line 16: width is alright. but let's practice and discover flex and how to work with it. let's use `flex-basis`

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  .block-3 {
+    order: 1;
+  }
+
+  .block-5,
+  .block-6 {
+    order: 2;
+  }
```
ihor-jpeg, line 37: 👍

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+@media (min-width: 900px) {
+  .block {
+    width: 33.33%;
+  }
```
ihor-jpeg, line 43: 33.33 * 3 is 99.99, not 100

Screenshot 2022-03-22 at 16 11 55
Screenshot 2022-03-22 at 16 12 02

it's a minor thing, but you should be aware of that. any idea how to make it 33.3333.....%?
also, because we are learning flex, let's use `flex-basis`

ihor-jpeg (changes requested):
Very close!

ihor-jpeg (approved):
Well done! 🚀

PR #1273, opened 2022-03-23
snake

PRO-GRAM-MER:
- [DEMO LINK](https://PRO-GRAM-MER.github.io/layout_snake/)
- [TEST REPORT LINK](https://PRO-GRAM-MER.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
+  color: #fff;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  flex-basis: 33%;
+  flex-grow: 1;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  .block-4 {
+    order: 3;
+  }
+
+  .block-6 {
+    background-color: #000;
+    order: 6;
+  }
```
ihor-jpeg, line 68: why repeat the smae style? (bg-color)

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+  .block-4 {
+    order: 4;
+  }
+
+  .block-6 {
+    background-color: black;
+    order: 6;
+  }
```
ihor-jpeg, line 86: same here. why repeat the same color?

ihor-jpeg (approved):
Good job! Have a look at my comment and let's remove redundant styles

src/style.css
PRO-GRAM-MER: modification implemented

PR #1275, opened 2022-03-23
snake number

Sardorbek1122:
- [DEMO LINK](https://Sardorbek1122.github.io/layout_snake/)
- [TEST REPORT LINK](https://Sardorbek1122.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,91 @@
+body {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.container-block {
+  display: flex;
+  flex-wrap: wrap;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,91 @@
...
+
+.block-6 {
+  background-color: rgb(0, 0, 0);
+  order: 6;
+}
+
+@media (min-width: 900px) {
+  .container {
+    min-width: 30vw;
+  }
```
ihor-jpeg, line 59: it doesn't seem like these styles are used in html. why do you need them?

ihor-jpeg (approved):
Nice! Take a look at my coment.

PR #1277, opened 2022-03-23
dysfunctional snake

WojciechCzarnowski:
- [DEMO LINK](https://WojciechCzarnowski.github.io/layout_snake/)
- [TEST REPORT LINK](https://WojciechCzarnowski.github.io/layout_snake/report/html_report/)

Stretching past 1200px creates problems

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,79 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  flex-grow: 1;
+  flex-shrink: 0;
+
+  height: 300px;
+  flex-basis: 300px;
+
+  justify-content: center;
+  align-items: center;
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,80 @@
...
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  flex-grow: 1;
+  flex-shrink: 0;
```
ihor-jpeg, line 13: what is the default of `flex-shrink`?

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+  color: white;
+}
+
+.box--1 {
+  background-color: red;
+}
+
+.box--2 {
+  background-color: rgb(204, 0, 0);
+}
```
ihor-jpeg, line 32: be consistent with this. do not use both RGB and color names. I'd only use RGB or hex because it's cooler

ihor-jpeg (approved):
Well done! Take a look at the comments

PR #1301, opened 2022-04-01
Snake is built

enrikadej:
- [DEMO LINK](https://enrikadej.github.io/layout_snake/)
- [TEST REPORT LINK](https://enrikadej.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,84 @@
+.html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-weight: 100;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,73 @@
+.html {
+  font-family: Arial, Helvetica, sans-serif;
+}
```
Anna-Tkachenko, line 3:
```suggestion
html {
  font-family: Arial, Helvetica, sans-serif;
}
```

Anna-Tkachenko (approved):
Good job!
Just fix font-family for text in your blocks and redeploy test report link

PR #1322, opened 2022-04-18
Solution

ABfeMate:
- [DEMO LINK](https://ABfeMate.github.io/layout_snake/)
- [TEST REPORT LINK](https://ABfeMate.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,89 @@
...
+  background-color: #600;
+}
+
+.block:nth-child(5) {
+  background-color: #300;
+}
+
+.block:nth-child(6) {
+  background-color: #000;
+}
```
Bohdan-Maksymiuk, line 47: you can use loop here instead of hardcoding

Bohdan-Maksymiuk (changes requested):
looks good but fix demo links - they don't work
maybe you forgot run deploy?

src/style.css
ABfeMate: I don't know how to use loops in css, so I'll leave it for later.

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1323, opened 2022-04-18
Develop

Havrysh:


src/style.css
```diff
@@ -0,0 +1,88 @@
...
+    order: 4;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    width: 33.33%;
+  }
+  .block-1 {
+    order: -1;
```
ArtemTopolskyi, line 71: don't use negative orders

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+.block-5 {
+  background-color: rgb(51, 0, 0);
+}
+
+.block-6 {
+  background-color: black;
+}
+
+@media (min-width:600px) and (max-width: 900px) {
+  .block-1 {
```
ArtemTopolskyi, line 46: same

ArtemTopolskyi (changes requested):

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  }
+  .block-4 {
+    order: 5;
+  }
+  .block-5 {
+    order: 4;
+  }
+  .block-6 {
+    order: 3;
+  }
```
Bohdan-Maksymiuk, line 87: do you really need to hardcode order for every block?

Bohdan-Maksymiuk (changes requested):

src/style.css
Havrysh: not really)

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1326, opened 2022-04-22
Develop

frontend112:
https://frontend112.github.io/layout_snake/
https://frontend112.github.io/layout_snake/report/html_report/


src/style.css
```diff
@@ -0,0 +1,90 @@
...
+
+  .block--3 {
+    order: 4;
+  }
+}
+
+@media only screen and (min-width: 769px) {
+  .block {
+    max-width: none;
+    flex-basis: 30%;
```
meforoma, line 76: umm, it actually should be 1/3rd so not actually 30% 🙂 
```suggestion
    flex-basis: calc(100% / 3);
```

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  background-color: rgb(51, 0, 0);
+  order: 5;
+}
+
+.block--6 {
+  background-color: rgb(0, 0, 0);
+  order: 6;
+}
+
+@media only screen and (min-width: 481px) {
```
meforoma, line 55: check your breakpoints (i guess 600 and 900px)
and you have these when reaching `one column layout`:
Snake

meforoma (changes requested):

meforoma (approved):

PR #1351, opened 2022-05-04
all test passed

Galyopa:
- [DEMO LINK](https://galyopa.github.io/layout_snake/)
- [TEST REPORT LINK](https://galyopa.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,79 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  font-weight: 400;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.box {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 300px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  background-color: rgb(20%, 0%, 0%);
+  order: 2;
+}
+
+.box-6 {
+  order: 1;
+  background-color: rgb(0%, 0%, 0%);
+}
+
+@media (max-width: 899px) {
```
Olexandr6, line 53: Use `min width` instead
```suggestion
@media (max-width: 899px) {
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+    order: 2;
+  }
+
+  .box-5,
+  .box-6 {
+    order: 3;
+  }
+}
+
+@media (max-width: 599px) {
```
Olexandr6, line 72: Here use `min-width`, as well. In such way it will be easier to read

Olexandr6 (changes requested):
Great job🔥 But some things can be improved; check comments👇

PR #1357, opened 2022-05-05
zig zag order

NikitaKarbazin:
- [DEMO LINK](https://NikitaKarbazin.github.io/layout_snake/)
- [TEST REPORT LINK](https://NikitaKarbazin.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -9,13 +9,13 @@ body {
 .flexbox {
   display: flex;
   flex-direction: column;
-  width: 100%;
 }
 
 .box {
   display: flex;
   height: 300px;
-  min-width: 300px;
+  flex-basis: 300px;
+  flex-grow: 1;
   background-color: red;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,140 @@
+body {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
```
Olexandr6, line 5: Always set fonts in the `html` selector. It's better approach

src/style.css
```diff
@@ -0,0 +1,140 @@
...
+  .box-4 {
+    order: 5;
+  }
+  .box-5 {
+    order: 4;
+  }
+  .box-6 {
+    order: 3;
+  }
+}
```
Olexandr6, line 140: Omg, you don't need such a count of media queries in this task. Try to improve this, using only 2 media queries🙂

G-YaSha (changes requested):
Your solution must be improved, check comments👇

G-YaSha (approved):
Looks much better🔥

PR #1358, opened 2022-05-05
Small fixes

yanaost:
- [DEMO LINK](https://yanaost.github.io/layout_snake/)
- [TEST REPORT LINK](https://yanaost.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,82 @@
+body {
+  margin: 0;
+}
+
+.block {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block * {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  text-align: center;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  flex-grow: 1;
+  flex-basis: 33%;
+}
+
+.block1 {
+  background: rgb(255,0,0);
+}
+
+.block2 {
+  background: rgb(calc(255 * 0.8),0,0);
```
mykola-koval, line 29: you can use percentage here. And use background-color instead of background shorthand.
https://developer.mozilla.org/en-US/docs/Web/CSS/background
```suggestion
  background-color: rgb(80%, 0%, 0%);
```

mykola-koval (approved):
👍

PR #1371, opened 2022-05-06
reading code html and css

Prince3378:
- [DEMO LINK](https://Prince3378.github.io/layout_snake/)
- [TEST REPORT LINK](https://Prince3378.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,103 @@
+body {
+  margin: 0;
+}
+
+.blok {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.blokk {
+  display: flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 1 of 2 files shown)


illiarerun (changes requested):
Nice start, now let's work on fixes:

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  .blok3 {
+    order: 3;
+  }
+  .blok4 {
+    order: 6;
+  }
+  .blok6 {
+    order: 3;
+  }
+}
```
illiarerun, line 64: you don't need these rule at all, just move these style to 900px rule
```suggestion
```

illiarerun (changes requested):

PR #1382, opened 2022-05-08
Created snake task

olya-yevsieienko:
- [DEMO LINK](https://olya-yevsieienko.github.io/layout_snake/)
- [TEST REPORT LINK](https://olya-yevsieienko.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  .block--5 {
+    order: 3;
+  }
+
+  .block--6 {
+    order: 4;
+  }
+}
+
+@media (max-width: 599px) {
```
jstmpelowycz, line 75: That is more obvious to create a media query this way
```suggestion
@media (min-width: 600px) {
```

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  background-color: rgb(51, 0, 0);
+  order: 2;
+}
+
+.block--6 {
+  background-color: rgb(0, 0, 0);
+  order: 1;
+}
+
+@media (max-width: 899px) {
```
jstmpelowycz, line 53: That is more obvious to create a media query this way
```suggestion
@media (min-width: 900px) {
```

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-grow: 1;
+  flex-basis: 33.3%;
```
jstmpelowycz, line 21:
```suggestion
  flex-basis: calc(100% / 3);
```

jstmpelowycz (changes requested):
A few things left to fix, and you're done

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
```
jstmpelowycz, line 16: The width of each block should be not less than its height (300px)

jstmpelowycz, line 16: The width of each block should be not less than its height (300px)

jstmpelowycz (changes requested):

src/style.css
jstmpelowycz: Add `min-width`

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  .block--4 {
+    order: 2;
+  }
+
+  .block--5 {
+    order: 1;
+  }
+}
+
+@media (min-width: 600px) and (max-width: 899px) {
```
jstmpelowycz, line 64: Do you need `max-width`?

jstmpelowycz (changes requested):
Small hints left

olya-yevsieienko:
I have a question, if I set flex-basic for each block, do I still need to
write min-width?

пт, 13 мая 2022 г. в 10:31, Dmytro Pelovych ***@***.***>:

olya-yevsieienko:
Okay, thanks! I fixed it.

ср, 11 мая 2022 г. в 20:14, Dmytro Pelovych ***@***.***>:

PR #1386, opened 2022-05-08
Snake has already crawled into your page

Andriy3323:
- [DEMO LINK](https://andriy3323.github.io/layout_snake/)
- [TEST REPORT LINK](https://andriy3323.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  min-width: 300px;
+  line-height: 300px;
+  text-align: center;
+  flex-basis: 33.33%;
+  flex-grow: 1;
+}
+
+.block__1 {
+  order: 1;
+  background-color: rgb(255 0 0);
```
illiarerun, line 29: use commas to separate values, it'll improve readability
```suggestion
  background-color: rgb(255, 0, 0);
```

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  height: var(--block_height);
+  min-width: 300px;
+  line-height: 300px;
+  text-align: center;
+  flex-basis: 33.33%;
+  flex-grow: 1;
+}
+
+.block__1 {
+  order: 1;
```
illiarerun, line 28: why do you need these custom orders by default? remove them and change them only in media rules

illiarerun (changes requested):
GJ, few fixes left:

PR #1388, opened 2022-05-08
Snake - task complete

IvanGrabovsky:
- [DEMO LINK](https://IvanGrabovsky.github.io/layout_snake/)
- [TEST REPORT LINK](https://IvanGrabovsky.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  background-color: #000;
+}
+
+.color-block {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  text-align: center;
+  line-height: 300px;
+  width: 33.3%;
+  color: #fff;
+  background-color: #f00;
+  /* height: 300px; */
+}
+
+.block--2 {
+  background-color: rgba(255, 0, 0, 0.8);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.color-block {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  text-align: center;
+  line-height: 300px;
+  width: 33.3%;
+  color: #fff;
+  background-color: #f00;
+  /* height: 300px; */
```
illiarerun, line 21: don't leave commented code
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,78 @@
+* {
+  margin: 0;
+  padding: 0;
```
illiarerun, line 3: don't reset your margins and paddings with `*` selector, it's a bad practice. Better use tag names for that
```suggestion
```

illiarerun (changes requested):
GJ, few fixes left todo. Also, my advice would be to choose mobile-first styling, that way you'd have much less CSS code

src/style.css
IvanGrabovsky: done

src/style.css
IvanGrabovsky: > GJ, few fixes left todo. Also, my advice would be to choose mobile-first styling, that way you'd have much less CSS code

done

illiarerun (approved):

PR #1396, opened 2022-05-09
add task

averstiuk:
- [DEMO LINK](https://averstiuk.github.io/layout_snake/)
- [TEST REPORT LINK](https://averstiuk.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,27 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+
+  <body>
+    <div class="blocks">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  .block-6 {
+    order: 1;
+  }
+}
+
+@media (min-width:1200px) {
+  .blocks {
+    max-width: 1199px;
+  }
+}
```
illiarerun, line 79: don't hardcode conatiner width. there should be 3 blocks on any screen width > 900px. try to find a solution, using a `flex-basis` style. now your solution looks like this
image
```suggestion
```

illiarerun (changes requested):
Nice start, but you need to fix this:

src/index.html
```diff
@@ -1,14 +1,27 @@
...
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
jstmpelowycz, line 6: Format your code properly

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  font-size: 100px;
+  color: white;
+}
+
+.blocks {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
```
jstmpelowycz, line 13: The width of each block should be not less than its height (300px)

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  .block-6 {
+    order: 3;
+  }
+}
+
+@media (min-width:900px) {
+  .block {
+    flex-basis: 33.3%;
```
jstmpelowycz, line 65:
```suggestion
    flex-basis: calc(100% / 3);
```

jstmpelowycz (changes requested):
A few things to fix, and you're done

src/style.css
jstmpelowycz: You forgot to fix it

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1403, opened 2022-05-09
Solution

ostapfriend:
- [DEMO LINK](https://ostapfriend.github.io/layout_snake/)
- [TEST REPORT LINK](https://ostapfriend.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,81 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  flex-grow: 1;
+  flex-basis: 300px;
+  color: #fff;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  background-color: red;
+}
+
+/*#region Background*/
+.block-2 {
+  background-color: rgb(80%, 0%, 0%);
+}
+
+.block-3 {
+  background-color: rgb(60%, 0%, 0%);
+}
+
+.block-4 {
+  background-color: rgb(40%, 0%, 0%);
+}
+
+.block-5 {
+  background-color: rgb(20%, 0%, 0%);
+}
+
+.block-6 {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  background-color: rgb(20%, 0%, 0%);
+}
+
+.block-6 {
+  background-color: black;
+}
+/*#endregion*/
+
+/*#region mediaWidth*/
+@media (min-width:600px) and (max-width:750px) {
```
illiarerun, line 46: that what i see between 750 and 900 px
image
remove this max-width par of media rule, why do you need it?
```suggestion
@media (min-width:600px) {
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+}
+
+@media (min-width:900px) and (max-width:1200px) {
+  .block {
+    flex-basis: 30%;
+  }
+  .block-1,
+  .block-2,
```
illiarerun, line 64: these blocks have `order` with a value of 0, so they will be upfront by default
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,81 @@
...
+  .block-5 {
+    order: 2;
+  }
+  .block-6 {
+    order: 3;
+  }
+
+}
+
+@media (min-width:900px) and (max-width:1200px) {
```
illiarerun, line 59: that's what I see on widths > 1200
image
remove this max-width part too

illiarerun (changes requested):
Nice start, now let's work on fixes:

src/style.css
```diff
@@ -0,0 +1,79 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
```
jstmpelowycz, line 10: The width of each block should be not less than its height (300px)

jstmpelowycz (changes requested):
Screenshot 2022-05-11 at 20 40 12
That's what I see if open your demo, that's not how it should be, try to fix your media queries

PR #1419, opened 2022-05-11
snake-block

Ihor-Novakovskyi:
- [DEMO LINK](https://Ihor-Novakovskyi.github.io/layout_snake/)
- [TEST REPORT LINK](https://Ihor-Novakovskyi.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,85 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  color: white;
+  max-width: 1199px;
+}
+
+.block {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-grow: 1;
+  flex-shrink: 0;
+  flex-basis: 300px;
+  height: 300px;
+  font-size: 100px;
+}
+
+.block-1 {
+  background-color: red;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,85 @@
...
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  color: white;
+  max-width: 1199px;
```
jstmpelowycz, line 13: You don't need it

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  background-color: rgb(20%, 0%, 0%);
+  order: 2;
+}
+
+.block-6 {
+  background-color: black;
+  order: 1;
+}
+
+@media (max-width: 899px) {
```
jstmpelowycz, line 54: That's more obvious
```suggestion
@media (min-width: 900px) {
```

src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  }
+  .block-5 {
+    order: 3;
+  }
+  .block-6 {
+    order: 4;
+  }
+}
+
+@media (max-width: 599px) {
```
jstmpelowycz, line 69: That's more obvious
```suggestion
@media (min-width: 600px) {
```

jstmpelowycz (changes requested):
A few things to fix, and you're good to go further

PR #1420, opened 2022-05-11
Working

illia-kyselov:


src/index.html
```diff
@@ -1,14 +1,24 @@
...
+  </head>
+  <body>
+    <div class="field">
+      <div class="field__block_1 block">1</div>
+      <div class="field__block_2 block">2</div>
+      <div class="field__block_3 block">3</div>
+      <div class="field__block_4 block">4</div>
+      <div class="field__block_5 block">5</div>
+      <div class="field__block_6 block">6</div>
+    </div>
```
marina-tilniak, line 22:
```suggestion
    <div class="field">
      <div class="block">1</div>
      <div class="block">2</div>
      <div class="block">3</div>
      <div class="block">4</div>
      <div class="block">5</div>
      <div class="block">6</div>
    </div>
```

marina-tilniak: use nth-child

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  background-color: rgb(48, 0, 0);
+}
+
+.field__block_6 {
+  background-color: rgb(0, 0, 0);
+}
+@media (min-width: 300px) {
+  .field {
+    flex-basis: 100%;
+  }
```
marina-tilniak, line 51: these styles do nothing. delete it. makes sense to add flex-basis only in situation when you hame more than one block. Here you have only one field and its children

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  .block:nth-child(6) {
+    order: 6;
+  }
+}
+
+@media (min-width: 600px) {
+  .field {
+    flex-basis: 50%;
+  }
```
marina-tilniak, line 75: same here

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  .block:nth-child(4) {
+    order: 3;
+  }
+}
+
+@media (min-width: 900px) {
+  .field {
+    flex-basis: 33.33%;
+  }
```
marina-tilniak, line 87: same here

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  .block:nth-child(4) {
+    order: 4;
+  }
+  .block:nth-child(5) {
+    order: 5;
+  }
+  .block:nth-child(6) {
+    order: 6;
+  }
```
marina-tilniak, line 69: for this width you can leave default order

marina-tilniak (changes requested):

marina-tilniak (approved):

PR #1438, opened 2022-05-20
snake was created

DimaShm:


src/style.css
```diff
@@ -0,0 +1,142 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  --field-width: 100%;
+}
+
+@media (min-width: 900px) {
```
marina-tilniak, line 10: be consistent, use only max-width or min-width

PR #1478, opened 2022-06-07
Snake is ready

opanasiukvlad:
- [DEMO LINK](https://opanasiukvlad.github.io/layout_snake/)
- [TEST REPORT LINK](https://opanasiukvlad.github.io/layout_snake/report/html_report/)

I replaced ".container * " selector with detailed one because global selector didn't worked.
I guess the reason is because i changed order for 3 properties, not for all, and when you try to change it to ) for all child's, it doesn't work.


VETALEK (approved):
Like 👍

PR #1480, opened 2022-06-07
added snake

silchyn:
- [DEMO LINK](https://silchyn.github.io/layout_snake/)
- [TEST REPORT LINK](https://silchyn.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -99,7 +99,7 @@
 
 @media (min-width: 900px) {
   .block {
-    flex-basis: 33vw;
+    flex-basis: 33%;
   }
 
   .block--1 {
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,4 @@
-body {
+.body {
   margin: 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,82 @@
+body {
+  margin: 0;
+}
```
VETALEK, line 3: better use some class, attached to `body`😉

src/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: 33vw;
+  }
+
+  .block--3 {
+    order: 0;
+  }
```
VETALEK, line 69: try to specify the `order` for each block, because it doesn't on my screen 🥲
image

VETALEK (changes requested):
commented 🔥

src/style.css
```diff
@@ -0,0 +1,128 @@
...
+
+  .block--6 {
+    order: 5;
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    flex-basis: 33vw;
```
VETALEK, line 102:
```suggestion
    flex-basis: 33%;
```
when you use `vw` it brokes, because of page scroll 😉

VETALEK (changes requested):
Almost 💪

VETALEK (approved):
Fire 🔥

PR #1496, opened 2022-06-08
layout_snake

viktoria-krasnyuk:
[DEMO LINK](https://viktoria-krasnyuk.github.io/layout_snake/)
[TEST REPORT LINK](https://viktoria-krasnyuk.github.io/layout_snake/report/html_report/)


marina-tilniak (changes requested):
use only max-width or min-width and delete extra flex-basis

Bohdan-Maksymiuk (approved):

PR #1512, opened 2022-06-09
snake done

Vlad-Fedorishchev:
https://FedorishchevvV.github.io/layout_snake/

https://FedorishchevvV.github.io/layout_snake/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,73 @@
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,73 @@
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
```
VETALEK, line 4: better move typography styles to the `html{}` 😉

VETALEK (approved):
Like 👍

PR #1517, opened 2022-06-09
fix first test

ruslan-petrenko:


src/style.css
```diff
@@ -0,0 +1,109 @@
+body {
+  margin: 0;
+}
+
+.container {
+  padding: 1px;
```
VETALEK, line 6: why? 😭

VETALEK (changes requested):
Check your media queries 😉

image

image

src/style.css
ruslan-petrenko: otherwise the test does not want to pass 
but I will try to fix it 😄

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+
+  .block6 {
+    order: 2;
+
+  }
+}
+
+@media (min-width: 900px) {
+  .block {
+    width: 33.3%;
```
solaryasha, line 67: Use `flex-basis` for this purpose

solaryasha (changes requested):

src/style.css
ruslan-petrenko: done

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+@media (min-width: 600px) {
+  .block3 {
+    order: 1;
+  }
+
+  .block5 {
+    order: 2;
+
+  }
+
```
solaryasha, line 58:
```suggestion
  .block5 {
    order: 2;
  }

```

Не залишай пустий рядок в CSS правилах

solaryasha (changes requested):

solaryasha (approved):
Випадково реджектнув :)

src/style.css
ruslan-petrenko: прийняв :))))

PR #1519, opened 2022-06-10
layout_snake

yuriiLevchenko2022:
- [DEMO LINK](https://yuriiLevchenko2022.github.io/layout_snake/)
- [TEST REPORT LINK](https://yuriiLevchenko2022.github.io/layout_snake/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,96 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  font-size: 100px;
```
VETALEK, line 4: better move typography styles to the `html{}` 😉

VETALEK (approved):
Like 👍

src/style.css
yuriiLevchenko2022: Yep. You're right. My mistake.

PR #1530, opened 2022-06-13
Develop

Hennadii-Turovskyi:


aliceblunt8 (changes requested):
there should only 3 columns on big screens

Screenshot 2022-06-15 at 15 54 07

ArtemTopolskyi (approved):

PR #1537, opened 2022-06-15
Initial commit

Vasyliy22:
https://Vasyliy22.github.io/layout_snake/
https:Vasyliy22.github.io/layout_snake/report/html_report/


src/style.css
```diff
@@ -0,0 +1,82 @@
...
+    order: 3;
+  }
+  .block--4 {
+    order: 4;
+  }
+}
+
+@media (min-width: 1200px) {
+  .block {
+    width: 320px;
```
Bohdan-Maksymiuk, line 80: 🤨

Bohdan-Maksymiuk: do not use such hardcode

Bohdan-Maksymiuk (dismissed):
https://user-images.githubusercontent.com/94289038/174272145-e7c5de62-4b93-4b11-8c6a-8301d9bfee87.mov

there should be max 3 columns
+ you may use flex-basis 😉

.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
```
vkryvytskyy, line 2: image
Blocks shouldn't be less than 300px wide

vkryvytskyy (changes requested):

src/index.html
```diff
@@ -1,14 +1,21 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="container">
+      <div class="block block--1">1</div>
```
vkryvytskyy, line 13:
```suggestion
      <div class="block">1</div>
```
use `nth-child` instead of adding separate class

vkryvytskyy (approved):

PR #1568, opened 2022-07-03
Snake was added

Sergey-Mironenko:


src/style.css
```diff
@@ -0,0 +1,78 @@
+body {
+  margin: 0;
+}
+
+.row {
+  font-size: 100px;
+  font-family: Arial, Helvetica, sans-serif;
```
valerii-derkach, line 7: It is better to move `font-family` to `html `

valerii-derkach (approved):
Good! 👍

PR #1636, opened 2022-07-15
notSpecialSolution#1

Reptiloid044:
[DEMO LINK](https://Reptiloid044.github.io/layout_snake/)
[TEST REPORT LINK](https://Reptiloid044.github.io/layout_snake/report/html_report/)

Hi Andrii!

In this case, I am not pretty sure about passing the test, because I cannot see the test when I am doing <npm test>

So could you please tell me why I cannot see the test result with my index.html and your checklist? 
**Please help meee!**

When the problem will be solved, I will do some special effects for the snake=)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,124 @@
+body {
+  margin: 0;
+}
+
+@media (min-width: 900px) {
+  .block--1 {
+    order: 0;
+  }
+
+  .block--2 {
+    order: 1;
+  }
+
+  .block--3 {
+    order: 2;
+  }
+
+  .block--4 {
+    order: 3;
+  }
+
+  .block--5 {
+    order: 4;
+  }
+
+  .block--6 {
+    order: 5;
+  }
+}
+
+@media (max-width: 899px) {
+  .block--1 {
+    order: 0;
+  }
+
+  .block--2 {
+    order: 1;
+  }
+
+  .block--3 {
+    order: 2;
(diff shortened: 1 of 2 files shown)


Reptiloid044:
what should I do?

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex ;
```
AlinaMatuschak, line 11:
```suggestion
  display: flex;
```

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+    order: 4;
+  }
+  .block--6 {
+    order: 6;
+  }
+}
+
+@media (min-width: 900px) {
+  .block__item {
+    flex-basis: 33%;
```
AlinaMatuschak, line 50:
```suggestion
    flex-basis: calc(100% / 3);
```

src/style.css
```diff
@@ -0,0 +1,92 @@
...
+
+.block--3 {
+  background-color: #900;
+}
+
+.block--4 {
+  background-color: #600;
+}
+
+.block--5 {
```
AlinaMatuschak, line 86: Will be nice to use nth-child

AlinaMatuschak (approved):

PR #1807, opened 2022-08-17
solution

eugcheban:
- [DEMO LINK](https://eugcheban.github.io/layout_snake/)
- [TEST REPORT LINK](https://eugcheban.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,93 @@
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+}
+
+body {
+  margin: 0;
+}
+
+.box {
+  display: flex;
+}
+
+.block {
+  display: flex;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  align-items: center;
+  justify-content: center;
+  height: 300px;
+  min-width: 300px;
+  color: white;
+}
+
+.block-1 {
+  background-color: rgb(255, 0, 0);
+  order: 1;
```
OleksiiBlezniuk, line 25: No need to set order for default state, they are in right order by default

OleksiiBlezniuk (changes requested):
Your blocks are a bit smaller then they should be (on large screens) Have you tried to make them flexible? I mean they should become larger if there is a free space in container 🙂

PR #2114, opened 2022-10-30
Develop

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/layout_snake/)
[TEST REPORT LINK](https://lefteris-l.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -13,7 +13,7 @@ body {
   align-items: center;
   justify-content: center;
   height: 300px;
-  font-family: Arial, sans-serif;
+  font-family: Arial, Helvetica, sans-serif;
   font-size: 100px;
   color: #fff;
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  .block4 {
+    order: 3;
+  }
+
+  .block5 {
+    order: 2;
+  }
+}
```
mgrinko, line 76: It it better to put media together with the selector it affects not to search through the whole document looking for applied rules

mgrinko (approved):

src/style.css
lefteris-L: Noted. Thanks @mgrinko . Would it be better if I wrote the .block$ selectors as .block--$ (modifier)?

PR #2159, opened 2022-11-03
Snake First Attempt

Kavster92:
- [DEMO LINK](https://Kavster92.github.io/layout_snake/)
- [TEST REPORT LINK](https://Kavster92.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,115 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+  text-align: center;
+  line-height: 300px;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.blocks {
+  display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,115 @@
...
+
+.b5 {
+  background-color: rgb(51, 0, 0);
+}
+
+.b6 {
+  background-color: rgb(0, 0, 0);
+}
+
+@media (min-width: 600px) {
```
naziamoff, line 54: you can write the initial order by default, and in media just change the ones that need to be changed. This way you will need to write much less code

naziamoff (approved):
Great job! 🔥 
There's one thing to improve though

src/style.css
Kavster92: Ahh so you only need to specify the order for the exact ones that need to be shifted? Not all the blocks you are trying to move? ie. Once you specify the order of some blocks the rest file in according to the order set by the rules placed on the individual blocks?

PR #2672, opened 2023-02-23
made snake

KosolapovAnatolii:
- [DEMO LINK](https://KosolapovAnatolii.github.io/layout_snake/)
- [TEST REPORT LINK](https://KosolapovAnatolii.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,117 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  color: #fff;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+    order: 6;
+  }
+
+  .block--5 {
+    order: 5;
+  }
+
+  .block--6 {
+    order: 4;
+  }
```
liliia-livandovskaya, line 92: in media you need to add styles that are different from default

liliia-livandovskaya (approved):

PR #2744, opened 2023-03-07
Develop

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_snake/)
[TEST REPORT LINK](https://Zibi95.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+  --block-height: 300px;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: calc(var(--block-height * 3));
+}
+
+.block {
+  height: var(--block-height);
+  flex-basis: var(--block-height);
+  line-height: var(--block-height);
+  text-align: center;
+  flex-grow: 1;
+}
+
+.block-1 {
+  background-color: rgb(100%, 0%, 0%);
+}
+
+.block-2 {
+  background-color: rgb(80%, 0%, 0%);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
...
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: calc(var(--block-height * 3));
+}
+
+.block {
+  height: var(--block-height);
+  flex-basis: var(--block-height);
+  line-height: var(--block-height);
```
DorotaLeniecDev, line 22: To center block content vertically and horizontally you could use flex here as well, instead line 22 and 23 you could write 
`
    display: flex;
    align-items: center;
    justify-content: center;
`

DorotaLeniecDev:
I can see that one of the point is not finalized 

- Blocks must be put in 2 to 6 rows and stretched: each row must fit either 1, 2, or 3 blocks. See the screenshots

Please double check this :)

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: #fff;
+  --block-height: 300px;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: calc(var(--block-height * 3));
```
DorotaLeniecDev, line 16: This is not a valid CSS property.  I think you wanted to set min-width so be 300px right?

DorotaLeniecDev (changes requested):
Small adjustmens needed.

src/style.css
Zibi95: Yea, but for one liners content this makes one property less. Should i change that?

DorotaLeniecDev (changes requested):
One last change and we are good to go :)

DorotaLeniecDev (approved):
Good job :)

PR #2921, opened 2023-04-20
add task solution

KatyaZaremska:
- [DEMO LINK](https://KatyaZaremska.github.io/layout_snake/)
- [TEST REPORT LINK](https://KatyaZaremska.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,87 @@
+body {
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 100px;
+  color: white;
+}
+
+.heading {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  margin: -1px;
+  border: 0;
+  padding: 0;
+  white-space: nowrap;
+  clip-path: inset(100%);
+  clip: rect(0 0 0 0);
+  overflow: hidden;
+}
+
+.wrapper {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.block {
+  flex-basis: 100%;
+  display: flex;
+  justify-content: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  width: 1px;
+  height: 1px;
+  margin: -1px;
+  border: 0;
+  padding: 0;
+  white-space: nowrap;
+  clip-path: inset(100%);
+  clip: rect(0 0 0 0);
+  overflow: hidden;
+}
```
naziamoff, line 19: what is that for?

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  justify-content: center;
+  align-items: center;
+  height: 300px;
+  min-width: 300px;
+  background-color: red;
+}
+
+.block--2 {
+  order: 2;
+  background-color: #d50000;
```
naziamoff, line 38: let's use rgb here instead, it will look much easier and clear in this case

src/style.css
```diff
@@ -0,0 +1,87 @@
...
+  order: 5;
+  background-color: #300;
+}
+
+.block--6 {
+  order: 6;
+  background-color: black;
+}
+
+@media (min-width: 600px) and (max-width: 899px) {
```
naziamoff, line 61: if you remove max-width here it will still work as expected

naziamoff (changes requested):

src/style.css
KatyaZaremska: I hide h1 with the name, forgot to delete it *facepalm*

naziamoff (approved):

PR #2928, opened 2023-04-20
task solution

julshvets:
- [DEMO LINK](https://julshvets.github.io/layout_snake/)
- [TEST REPORT LINK](https://julshvets.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,104 @@
+body {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, sans-serif;
+}
+
+.block-container {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+
+}
+
+.block {
+  display: flex;
+  height: 300px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,104 @@
...
+  margin: 0;
+  padding: 0;
+  font-family: Arial, sans-serif;
+}
+
+.block-container {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+
```
naziamoff, line 11:
```suggestion
```

naziamoff (changes requested):
looks good already 👍 
But now let's find a way to remove one media query and keep the functionality untouched

naziamoff (approved):
🔥

src/style.css
julshvets: Fixed.
Thank you!

PR #2931, opened 2023-04-20
Task solved

olehmahera:
- [DEMO LINK](https://olehmahera.github.io/layout_snake/)
- [TEST REPORT LINK](https://olehmahera.github.io/layout_snake/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,19 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Snake</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Snake</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,19 @@
...
+    <title>Snake</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="block block1">1</div>
+    <div class="block block2">2</div>
+    <div class="block block3">3</div>
+    <div class="block block4">4</div>
+    <div class="block block5">5</div>
+    <div class="block block6">6</div>
```
naziamoff, line 17: we'll learn that later, in BEM topic, but what you have here (block1, block2 etc.) are a modifiers. They slightly change the state of the element. Syntactically they should look like that:
```suggestion
    <div class="block block--1">1</div>
    <div class="block block--2">2</div>
    <div class="block block--3">3</div>
    <div class="block block--4">4</div>
    <div class="block block--5">5</div>
    <div class="block block--6">6</div>
```

src/style.css
```diff
@@ -0,0 +1,79 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: stretch;
+}
```
naziamoff, line 10: nope, let's style by class, not by tag

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+.block1 {
+  order: 1;
+  background-color: red;
+}
+
+.block2 {
+  order: 2;
+  background-color: #c00;
```
naziamoff, line 34: can be done easier, with rgb(...)

naziamoff (approved):
Not bad 👍 
See some extra suggestions

